PR #14, opened 2015-02-22
Allow filled and empty glyphs to be specified as parameters to barstring()

jiahao:
Also defaults solidglyph to the Unicode code point █

U+2588 FULL BLOCK


timholy:
That does seem like a good idea.

PR #16, opened 2015-03-07
Only print final status message once, add finish! method to fast-forward progress meter

aaw:
Thanks for this great package! 

In one of my projects, I know approximately how many iterations a while loop is going to take but I often over-shoot or under-shoot the actual value by a little bit. This patch makes the output of ProgressMeter look sane in either case: when over-shooting, it won't print out the final "Progress: 1XX%" line more than once. When under-shooting, you can now call `finish!` to fast-forward the progress meter to 100% at the end of the loop. The tests I added give examples of both applications.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -2,7 +2,7 @@ module ProgressMeter
 
 using Compat
 
-export Progress, next!, cancel, @showprogress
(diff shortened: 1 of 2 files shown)


timholy:
Looks great, thanks very much!

PR #18, opened 2015-03-21
make the output stream selectable

bicycle1885:
This pull request makes the output stream of a progress selectable.
This is especially useful when we want to save an output of a program into a file with redirection and check the progress from our terminal (the `curl` command-line tool works as such):

```
$ julia program.jl > output.txt
# progress comes here
```

In this case, you can set the output stream of the progress bar to `STDERR`:

```julia
p = Progress(10)
p.output = STDERR
```

BTW, two constructors of `Progress` look a little bit messy and make it hard to set some parameters.
Can I unify them to a constructor using keyword arguments like `function Progress(n::Integer; dt::Real=1.0, desc::String="Progress: ", and others)` as another pull request?
Of course we will need some outer constructors like `Progress(n::Integer, desc::String)` for backward compatibility.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -14,8 +14,9 @@ type Progress
     desc::String     # prefix to the percentage, e.g.  "Computing..."
     barlen::Int      # progress bar size (default is available terminal width)
     color::Symbol    # default to green
+    output::IO       # output stream into which the progress is written
 
-    function Progress(n::Integer, dt::Real = 1.0, desc::String = "Progress: ", barlen::Int = 0, color::Symbol = :green)
+    function Progress(n::Integer, dt::Real = 1.0, desc::String = "Progress: ", barlen::Int = 0, color::Symbol = :green, output::IO = STDOUT)
         this = new(convert(Int, n), convert(Float64, dt), 0)
         this.tfirst = time()
         this.tlast = this.tfirst
@@ -23,6 +24,7 @@ type Progress
         this.desc = desc
         this.barlen = barlen
         this.color = color
+        this.output = output
         this
     end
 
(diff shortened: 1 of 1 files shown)


garrison:
I've always wondered if we might want to do more infrequent updates anytime we are not writing to a tty (e.g. write only when the progress meter actually updates, but then again this would not allow people to get the ETA until it got to 1%, so perhaps this is an imperfect idea).

This also reminds me: the current code looks at `tty_size()` if `barlen` is not given, but this is not going to be very relevant when writing to `STDOUT`.  Perhaps something arbitrary (e.g. 80 characters) should be chosen when `barlen` is not given and we are writing to something that is not a tty?

I generally like the idea of moving to use keyword arguments.  Unfortunately users of the `@showprogress` macro will need to continue using the old interface, as keyword arguments are not supported in macros as far as I can tell.

bicycle1885:
I don't have any strong opinion about the behavior of the progress meter.
But in my humble opinion, even if we mean well, some kind of "smart" progress meters will confuse users of this package.
People may want to see a progress meter that fits the width of their window even when the meter is written into a file. Nobody knows.
So I think that kind of decision should be left to package users.

Yes, I really have to care about compatibility when moving to use keyword arguments. Thank you for your advice.

bicycle1885:
This is a commit in which the constructor uses keyword arguments: 8f82f89fcf77b946797c2c9b926f5bb77e46d525

timholy:
I think when viewing a file, you should just be viewing what's in the file: at that point the connection to a "progress meter" is lost and it's just plain text. Anything else is going to depend on details of the viewer and be too hard (and likely, too error-prone) to maintain. So I agree with @garrison that a default needs to be considered.

It's also fine to combine the two constructors.

Thanks, @bicycle1885!

PR #19, opened 2015-03-23
WIP: unified constructor with keyword arguments

bicycle1885:
As I mentioned in #18, it would be worth unifying two inner constructors into an inner constructor using keyword arguments to make it easy to set some of the parameters and keeping backward compatibility; this is a working implementation of it.

But before finishing my job, let me ask a few questions.

1. Why the original two constructors specify different `dt` values (`1.0` vs. `0.01`)?
2. Again, why `barlen` is set to different values between the two constructors (`0` vs. tty width)?

If there is no reason to do so, I think setting a same value would be better.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
-julia 0.3-
+julia 0.3
 Compat
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6521" && __precompile__()
+
 module ProgressMeter
 
 using Compat


timholy:
Gracious me, I never noticed this. Sorry about that.

If you're still interested in this, it looks much nicer to me. To answer your questions:
- The different `dt` values probably come down from different contributions (3c23bf4c vs 9be39985 were the most recent to touch those lines). I can't think of a reason they shouldn't be synchronized, but CCing @abahm and @hofmannmartin.
- Re the `barlen`, also CCing @tlycken (see cda59ece).

bicycle1885:
Thank you for the reply. I almost forgot this pull request, but still I'm keen to finish it.
I'll take a look later and revise this one again.

hofmannmartin:
The different `dt` values probably origin in earlier contributions, as I only did changes related to the julia-0.3 -> julia-0.4 update.

tlycken:
Re: `barlen`, I also only worked on it to remove deprecation warnings going from 0.3 to 0.4, so I don't know which makes most sense.

bicycle1885:
Thank you! I think this is ready to review and to be merged. I've set the default `dt` value to `0.1` and kept `barlen` as is.

timholy:
Much nicer, many thanks!

PR #21, opened 2015-05-11
Added update! functionality

cbecker:
Hi Tim, this adds an `update!()` method which I found useful for certain situations, such as the one I added to the README.md file.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -14,8 +14,9 @@ type Progress
     desc::String     # prefix to the percentage, e.g.  "Computing..."
     barlen::Int      # progress bar size (default is available terminal width)
     color::Symbol    # default to green
+    output::IO       # output stream into which the progress is written
 
-    function Progress(n::Integer, dt::Real = 1.0, desc::String = "Progress: ", barlen::Int = 0, color::Symbol = :green)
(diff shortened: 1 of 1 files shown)


src/ProgressMeter.jl
```diff
@@ -70,11 +70,28 @@ function next!(p::Progress)
     end
 end
 
+function next!(p::Progress)
```
timholy: Now calling just `next!` will never update the progress bar, meaning that it will break existing code.

cbecker: oh sorry you are right, I forgot to fix this.

cbecker: Fixed now, see below.

timholy:
Ah, there's also a conflict with current master. You were probably behind when you implemented this---if the package maintainer (me) doesn't tag after each commit, then you can be modifying an outdated package. (You can protect yourself against that by checking out `master` first.)

Sorry about that. I just tagged a new release---it was definitely overdue. I made it a minor release, to reflect all the good work folks have been putting into this.

Anyway, you might need to do a `git fetch; git rebase -i origin/master`. More info is [here](http://docs.julialang.org/en/latest/manual/packages/#squashing-and-rebasing).

PR #22, opened 2015-05-11
Added update! method for non-constant progress steps

cbecker:
Should be clean now ;)

Tests pass as well.


timholy:
Looks great, many thanks!

garrison:
@cbecker Thanks!  I think this new feature should have some tests.  Would you be willing to add some (in a separate pull request)?

timholy:
Good point, @garrison. (I was so excited about the great docs that went along with it, I just didn't notice! :smile:)

PR #25, opened 2015-06-05
Refactor @showprogress macro to support multidimensional loops

### LINKED ISSUE #23: Support multidimensional for loops
It would be convenient if ProgressMeter supported multidimensional for loops like:
```
@showprogress for i=1:10, j=1:10
    sleep(0.1)
end
```

garrison:
Closes #23.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -153,19 +153,34 @@ function showprogress_process_expr(node, metersym)
             ($finish!)($metersym)
             $node
         end
-    elseif node.head === :continue
-        # special handling for continue statement
-        quote
-            ($next!)($metersym)
-            $node
-        end
     elseif node.head === :for || node.head === :while
-        # do not process break and continue statements in inner loops
+        # do not process inner loops
+        #
+        # FIXME: do not process break and return statements in inner functions
+        # either
         node
     else
         # process each subexpression recursively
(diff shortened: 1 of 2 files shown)


src/ProgressMeter.jl
```diff
@@ -177,67 +192,66 @@ macro showprogress(args...)
...
-        assignidx = 3
+        outerassignidx = endof(loop.args)
         loopbodyidx = 2
-        is_comprehension = true
     else
         throw(ArgumentError("Final argument to @showprogress must be a for loop or comprehension."))
     end
 
-    newloop = Expr(loop.head, loop.args...)
+    loop = copy(loop)
```
timholy: Come to think of it, do we even need to make a copy?

timholy:
Awesome work, as usual. Merge at will.

src/ProgressMeter.jl
garrison: This is the main thing I wondered about as well.  Can macros modify the AST in-place before returning them, or should they be purely functional (i.e. non-mutating)?  Not knowing the answer I went with what I considered to be the safe route for now.  If you don't know for sure, I'll ask on julia-users.

timholy: I assume it's OK since the typical usage is for them to come from parsing a text file. But to clarify, I don't have strong feelings one way or another.

PR #26, opened 2015-09-23
replaced String by AbstractString

hofmannmartin:
to fix "WARNING: Base.String is deprecated, use AbstractString instead." in Julia 0.4.0-rc2+10.

Pkg.test did pass on Julia 0.3.12-pre+5 and 0.4.0-rc2+10.


timholy:
Thanks!

PR #27, opened 2015-09-24
Use appropriate control character within ESS

dennisprangle:
This lets `printover` work as intended within ESS.

Note that one case where it still doesn't work is running `Pkg.test("ProgressMeter")` within ESS. However `include("runtests.jl")` works fine.


timholy:
Thanks!

PR #34, opened 2016-01-21
Add docstrings and a progress meter based on a threshold

timholy:
Some tasks, like minimizing a function, often take an indeterminate number of iterations: for example, one might iterate a L-BFGS step until the function value stops changing by more than `thresh`, and it's not predictable how many iterations this will take. This PR adds an alternative type of progress meter customized for such settings.

Example of output during the computation:
```
Computing nonnegative factorization: (thresh = 1e-07, value = 2.84137e-6)
```
Example of output once `value` drops below `thresh`:
```
Computing nonnegative factorization: Time: 0:00:09 (164 iterations)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.3
   - 0.4
   - nightly
 notifications:
--- a/README.md
+++ b/README.md
@@ -13,12 +13,26 @@ Pkg.add("ProgressMeter")
 
 ## Usage
 
+### Progress meters for tasks with a pre-determined number of steps
+
 This works for functions that process things in loops.
 Here's a demonstration of how to use it:
 
 ```julia
 using ProgressMeter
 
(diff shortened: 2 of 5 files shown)


timholy:
The test failure on julia 0.3 is due to the lack of support for `%g` in format strings. Should we stop supporting julia 0.3, or should I make it `%f`?

garrison:
My vote would be to stop supporting 0.3 (0.3 users can still use the newest released version that supports 0.3).  But I don't have strong feelings.

tlycken:
I would also vote for dropping 0.3. If anything significant has been added since the last tagged release, we should probably tag a release before merging this, too.

timholy:
OK, I turned off 0.3 support (will follow the tagging advice of @tlycken). I also updated the README, which includes moving some material around, like starting with `@showprogress` since that's probably the easiest way to use this package.

I'll wait a bit before hitting merge to give folks a chance to voice concerns.

Issue #35, opened 2016-03-07
barspec, e.g. "[=> ]"


zsunberg:
Hi all,

I have been extremely satisfied by all of the status bars crossing my screen since I discovered this package. However, sometimes the solid blocks just aren't what I'm looking for aesthetically. Would it be ok to allow customization of the characters in the progress bar?

The easiest way I can think of to do this is to add a keyword argument to the constructor called `barspec` that is a 5 character string. The first character is the left end, the second character is the bar fill character, the third character is the front end of the bar, the fourth character is the empty character, and the fifth character is the right end.

For example,
`"[#> ]"` would produce a bar like `[########>    ]`,
other examples could be "[=> ]", or "wot !" :)

Let me know if you think this is a good idea. I'll submit a PR in a few minutes.

zsunberg:
fixed by #37

PR #36, opened 2016-03-07
Pull request/9f3ab8d1

zsunberg:
Implements the idea in #35

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -256,12 +256,23 @@ end
 
 function barstring(barlen, percentage_complete; barspec::AbstractString="|██ |")
     bar = ""
-    glyphs = collect(graphemes(barspec))
-    leftglyph = glyphs[1]
-    solidglyph = glyphs[2]
-    frontglyph = glyphs[3]
-    emptyglyph = glyphs[4]
-    rightglyph = glyphs[5]
+    local leftglyph, solidglyph, frontglyph, emptyglyph, rightglyph
+    try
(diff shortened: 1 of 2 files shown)


tlycken:
I like the idea, but I'm less hot on the API.

If one has to specify all the characters anyway, maybe an `NTuple{5,Char}` is a better choice of argument type? That would make it more explicit to users exactly what input is expected (and we could leverage tuple unpacking before printing). There might very well be other possibilities as well; I'm just a little wary about accepting a string for something that is really a collection of exactly five `Char`s, each with a distinct purpose and meaning.

zsunberg:
Hmm... Yeah, I see your point @tlycken, there's a tradeoff between making it explicit and convenient. An `NTuple` doesn't seem too much more clear than a string to me, and it seems like overkill to have each character have its own argument.

Maybe the clearest option would be to make a little type for it with a field for each character, and that type could have a constructor that accepts 5 characters and a constructor that accepts a string
```julia
type BarChars
    leftend::Char
    fill::Char
    front::Char
    empty::Char
    rightend::Char
end
BarChars(s::AbstractString) = BarChars(collect(graphemes(s))...)
```

If ya'll think `NTuple` is the way to go, I'm totally fine with that as well. Let me know what you think and I'll update the PR

tlycken:
A `BarChars` type seems even better to me. This makes it explicit inside ProgressMeter.jl what's happening, and it makes it easy to give good error messages when the string is invalid, since that is easily isolated to the constructor of `BarChars`. Great suggestion!

PR #37, opened 2016-03-16
Custom Bar Glyphs

zsunberg:
Implements idea in #35 taking suggestions in #36 into account

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -256,12 +256,23 @@ end
 
 function barstring(barlen, percentage_complete; barspec::AbstractString="|██ |")
     bar = ""
-    glyphs = collect(graphemes(barspec))
-    leftglyph = glyphs[1]
-    solidglyph = glyphs[2]
-    frontglyph = glyphs[3]
-    emptyglyph = glyphs[4]
-    rightglyph = glyphs[5]
+    local leftglyph, solidglyph, frontglyph, emptyglyph, rightglyph
+    try
+        glyphs = collect(graphemes(barspec))
+        @assert length(glyphs) == 5
+        leftglyph = glyphs[1]
+        solidglyph = glyphs[2]
+        frontglyph = glyphs[3]
+        emptyglyph = glyphs[4]
+        rightglyph = glyphs[5]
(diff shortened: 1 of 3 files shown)


zsunberg:
oops - guess I could have just added commits to #36; didn't realize you could do that. Also, let me know if this looks good, and I'll update the README as well

src/ProgressMeter.jl
```diff
@@ -252,12 +281,20 @@ function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
     end
 end
 
-function barstring(barlen, percentage_complete; solidglyph="█", emptyglyph=" ")
+function barstring(barlen, percentage_complete; barglyphs::BarGlyphs=BarGlyphs('|','█','█',' ','|'))
```
tlycken: Minor style nitpick: here you use the `BarGlyphs(chars...)` constructor, while above (line 78) you use the `BarGlyphs(string)` constructor. Is there a reason not to use the same one?

test/test.jl
```diff
@@ -229,6 +230,20 @@ end
 println("Testing keyword arguments")
 testfunc13()
 
+function testfunc14(barglyphs)
+    n = 30
+    # full keyword argumetns
+    p = ProgressMeter.Progress(n, barglyphs=ProgressMeter.BarGlyphs(barglyphs))
```
tlycken: This is nice. It would be even nicer to add a test that uses this in conjunction with the other kwargs (I don't think it will make a difference now, but we should make sure that it doesn't in the future).

Also, a couple of tests that invoke both constructors of `BarGlyphs` would be nice, since they're part of the public interface.

tlycken:
Looks good overall to me! I added a couple of line comments with some minor things to address. Also, would you mind squashing the commits? (If you don't know how to, that's fine anyway.)

Thanks a lot for a great addition!

zsunberg:
Ok, addressed those comments and added a demo to the readme. To tired to learn about rebasing (I guess that's what squashing entails) tonight, but will read about it at some point.

tlycken:
LGTM, thanks!

PR #38, opened 2016-04-03
Fix deprecation: tty_size -> displaysize

Keno:


garrison:
This should have a `VERSION` check, since `Base.displaysize` does not exist in 0.4

Keno:
It does exist in Compat and is exported.

Keno:
Looks like the compat wrapper is broken. `displaysize` takes and argument while tty_size does not.

garrison:
In that case, I think you should call `displaysize`, not `Base.displaysize`, so that `Compat` can cover on julia 0.4.  But the compat wrapper will need fixed too, as you mention.

Keno:
Yes, you're right. I meant to do that. Sorry!

timholy:
Superseded by #39.

PR #39, opened 2016-04-22
Fix warnings on julia-0.5

timholy:


garrison:
The `displaysize` change overlaps with #38.  I don't believe [the issue with the Compat wrapper](https://github.com/timholy/ProgressMeter.jl/pull/38#issuecomment-204867733) has been fixed yet...

Issue #40, opened 2016-05-02
Allow update also of (status) values


robertfeldt:
Thanks for ProgressMeter; very useful. For long-running computations (such as for example in optimization) it would be nice to allow a dict (or some other way) to specify current values which can then be reported on the lines below the progress bar? For example, in the optimiation, when want to know the overall progress in terms of time (ETA) but also would like to know the objective value of the current best solution found. If these are just printed to STDOUT it would mess up the reporting from ProgressMeter so it seems there needs to be some support for this. Has this been considered as a future extension? If not, if this is something you could accept as a PR I might take a stab at it.

timholy:
Did you see the `ProgressThresh` type?

robertfeldt:
I did, yes, but my impression is that it is for cases when I want to affect the progress with said value. In my case I'm trying to find a way to present additional information about the progress but which should not determine the actual progress. Does it make sense? ;)

robertfeldt:
Example: At one point in an optimization run I may have done 50 out of 100 iterations and I want to present the current minima found say:

progress  50% ... ETA: 00:14:12
  current minima: 0.3452

It would be quite nice if it can also model the EVA (Estimated Value at Arrival) using information about the recent progress on the value or something. But most important is to be able to show additional information without messing up the progress meter output.

timholy:
It's intended for cases where convergence is determined by value rather than number of iterations---for example, an optimization in which convergence occurs when the function value changes by less than `tol`. It might take 3 or 3000 iterations to reach `tol`, and you can't predict this in advance. (Please do feel free to improve the documentation, given that it was not sufficiently clear.)

It looks like you have both a deterministic number of iterations and a value you want to report back. That's basically a hybrid of the two, so I'd guess that you'd indeed want a new type. I'd certainly consider a PR implementing this.

robertfeldt:
Ok, I have a busy week but will take a stab at this after that. Basic idea is for an optional argument to `next!` and `update!` like so:

```julia
p = ProgressMeter(100, 1, "progress  ", 70)
...
next!(p; minima = current_best.fitness, meanfitness = mean(map(i -> i.fitness, individuals))) # for some values of interest, here lets say minima is 0.25 and meanfitness is 0.567
````
which will print after 50 iterations:
````
progress  50% ||||||||||...bar here...|||||                     ETA: 00:07:13
  minima: 0.25
  meanfitness: 0.567
````
[...]

PR #41, opened 2016-05-05
Printing additional values below progress bar

robertfeldt:
This implements the feature request from https://github.com/timholy/ProgressMeter.jl/issues/40#issuecomment-216180088
I also added tests and there seems to be no regressions.


tlycken:
Sorry - my glitchy phone decided that scrolling was not fun enough...

robertfeldt:
@tlycken hehe, I was amazed my PR could be that bad it was rejected in a few seconds... ;)

timholy:
Very impressive, thanks a lot!

PR #47, opened 2016-06-16
Call srand() and rand() explicitly instead of isprime()

garrison:
Per discussion with @tlycken at https://github.com/timholy/ProgressMeter.jl/commit/ff64d214ac475e4eba2c96a4b4cba316cee7abf4#commitcomment-17892752


timholy:
Thanks!

PR #48, opened 2016-06-29
Fix hidden cursor after the progress with showvalues=

alyst:
The PR just removes the code that hides the cursor. An alternative would be to restore the cursor state after the progress is done, but it still would not handle `Ctrl+C` cases.


timholy:
I'm not a terminal expert, can you clarify what this fixes?

alyst:
@timholy well, after the progress bar finishes, the cursor just doesn't show up anymore in the whole julia/shell session, including the command prompt. (I'm using `konsole` terminal app)

timholy:
Oh, that does seem rather bad :smile:. Thanks!

Issue #60, opened 2016-10-26
STDOUT by default?


RedPointyJackson:
Why does this package uses `STDOUT` by default? It can be very cumbersome to manually redirect every progress bar if one wants to pipe the output of a program to a file.

alexmorley:
Seems to me like this would be a decent default as the most common use case for a progress bar would be displaying the output on STDOUT?

RedPointyJackson:
Why on STDOUT? For example, `wget` shows progress in STDERR, like `curl`. The only applications that use STDOUT are the ones not designed to be piped, which is very limiting for a general purpose language like Julia. 

Also, displaying the bar in STDERR has no effect whatsoever for people who does not intend to redirect the output, and is very convenient to people who does.

alexmorley:
For sure, I didn't realise that was the alternative you were proposing.

RedPointyJackson:
Oh, sorry, I should have been more clear in the issue.

timholy:
Want to submit a PR to change it?

timholy:
Closed by #64

PR #63, opened 2017-01-05
Change default output to stderr

alexmorley:
As discussed in #60


timholy:
Thanks! I think this makes sense, but I'll leave it open for a couple of days to see if any objections arise. If I forget to merge, don't hesitate to ping me.

PR #64, opened 2017-01-06
Change default output STDOUT by STDERR

RedPointyJackson:
As discussed in issue #60.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -50,7 +50,7 @@ end
 
 """
 `prog = Progress(n; dt=0.1, desc="Progress: ", color=:green,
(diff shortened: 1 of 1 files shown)


timholy:
Thanks for contributing this! A slight bummer is that we now have two, see #63. I guess it might be best to CC others who had been involved in the discussion (like this, @alexmorely). @alexmorely, I think I will merge this one first, since @RedPointyJackson is the one who discovered the issue. But since you caught an extra spot that needed changing, I'll merge yours second. Fairest approach I can come up with.

timholy:
Argh, I messed up the CC, sorry @alexmorley.

PR #65, opened 2017-03-01
Remove `@showprogress` tests for dict comprehensions on julia 0.6

garrison:
Dict comprehensions are gone in 0.6, replaced with a "generator" syntax. Since `@showprogress` does not support generators (yet), I've simply disabled these tests on julia 0.6.


garrison:
Travis will succeed once julia nightly there is updated to be 0.6.0-pre (not dev).

timholy:
Seems reasonable. Thanks!

PR #66, opened 2017-03-15
Fix Julia 0.6 deprecations

martinholters:
* abstract type declaration (requires Compat 0.17)
* inner constructor syntax

Note: Tests are still expected to fail on 0.6 pending #65.


timholy:
Thanks!

PR #72, opened 2017-06-13
Fix warnings regarding deprecation of (`immutable`) `type`

martinholters:
It's not pretty, but works...

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -23,16 +23,19 @@ ProgressMeter
 
 
 
+eval(quote
 """
 Holds the five characters that will be used to generate the progress bar.
 """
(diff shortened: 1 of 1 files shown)


timholy:
Is there no Compat solution for this?

martinholters:
Ref. JuliaLang/Compat.jl#338

timholy:
Hmm, interesting issues. I'd almost rather just require 0.6, how do you feel about that?

martinholters:
> It's not pretty

... so no offense taken if you rather not go with this.

timholy:
Closed by #81. Sorry for not taking this!

martinholters:
> Sorry for not taking this!

No worries!

PR #74, opened 2017-08-03
Fix example

IlyaOrson:
Clarifies a little the use of keywords instead of adding another method to fix the broken example


timholy:
Thanks!

PR #77, opened 2017-08-15
support color in IJulia, fix truncation

### LINKED ISSUE #76: Progress bar stops after a while in Jupyter notebooks
When the task is too long, Jupyter stops the progress bar with the message:

>Excessive output truncated after 524411 bytes.

According to https://github.com/JuliaLang/IJulia.jl/issues/455#issuecomment-322462631, we need to reset `IJulia.stdio_bytes[] = 0` after a while.

Where should this reset happen in the source code?

stevengj:
Closes #76.


timholy:
Thanks!

PR #82, opened 2017-10-09
added pmap to fix #32

### LINKED ISSUE #32: Working with pmap
I thought I should note here that I found working with pmap and ProgressMeter a bit tricky, so I bundled the solution as a very simple package: https://github.com/slundberg/PmapProgressMeter.jl which provides `pmap(p::Progress, f::Function, values)`.

This is related to #9 and it may be of interest to eventually merge this into ProgressMeter.

zsunberg:
Added the `pmap` function to fix #32

- Not sure if the global dictionaries are the right way to do it, open to other suggestions.
- showvalues is not implemented yet; not sure about the best way to do that if it is needed.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -487,4 +487,55 @@ macro showprogress(args...)
     end
 end
 
+# global meter references for pmap 
+const global_meters = Dict{String, AbstractProgress}()
+const global_progress = Dict{String, Int}()
+const global_print_lock = Dict{String, ReentrantLock}()
+
+"""
+    pmap(f::Function, p::ProgressMeter, c...; passcallback=false, kwargs...)
+
+`pmap` with a progress meter. If the keyword argument passcallback is true, then passes an additional first argument to the function which is a callback taking one argument, n, to add n to the progress meter. If it is false, then updates the progress meter by 1 every time a parallel call finishes.
+"""
+function Base.pmap(f::Function, p::Progress, values...; kwargs...)
+    global global_meters
+    global global_progress
+    global global_print_lock
+
+    id = randstring(50)
+    global_meters[id] = p
+    global_progress[id] = 0
+    global_print_lock[id] = ReentrantLock()
+
+    kwa = Dict(kwargs)
+    passcallback = get(kwa, :passcallback, false)
+    delete!(kwa, :passcallback)
+
+    out = pmap(values...; kwa...) do x...
+        if passcallback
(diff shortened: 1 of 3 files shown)


timholy:
Thanks for submitting this! Overall it looks good.

As you point out, the globals seem to be the main concern. I haven't written much parallel code in a while, so let me ask: do you need `global_print_lock` or would something specific to this particular `pmap` call be sufficient? If the latter, what about allocating these locally in `pmap` and passing them as arguments to a generic handler? I haven't thought this through in detail, so I can't promise this is sensible.

zsunberg:
Thanks for the suggestion. I will think about it when I get a chance. Also should look at https://github.com/jekyllstein/ParallelProgressMeter.jl/blob/master/src/ParallelProgressMeter.jl and https://github.com/adamslc/ProgressMeter.jl/blob/master/src/pmap.jl

zsunberg:
Tried for a while to figure out how to get all the plumbing for this in scope of the pmap function. My attempt is here: https://github.com/zsunberg/ProgressMeter.jl/blob/local_scope/src/ProgressMeter.jl#L495

Unfortunately it segfaults, and I don't have a good enough understanding of what is going on to fix it. Will think about how to reach out to others for help, but I'm out of time for working on it today.

zsunberg:
Sorry about letting this get stale :/ Unfortunately, I don't think I am going to have time to figure out how to do it better, and I don't like the way it is now with the global dictionaries, so feel free to close it. Someday someone will figure out how to do it right :)

timholy:
No problem. This is a great start, and fixing it up is on my TODO list. It's just that lots of other things are, too.

zsunberg:
Just in case this proves useful to you or someone who wants to finish this: another line of thinking besides remote calling back to process 1 would be to re-implement pmap with e.g. the simple implementation here https://docs.julialang.org/en/latest/manual/parallel-computing#Scheduling-1 and add an `update!(progress, ...)` line after the `remotecall_fetch` line.

But then this version of pmap would not have all the keyword argument functionality of the core implementation of pmap.

zsunberg:
Was thinking about this today. I think a good way to do it is to have one task that calls pmap and another task that handles the progress bar printing. The progress reports from the workers would be passed back to the printing task via channels. That seems like a much better solution than the remote calls that are currently in this PR. If I get a few hours soon, I think I can implement it. Or if anyone else wants to, they can throw it together.

PR #83, opened 2017-11-12
Replace control char \u1b[A with \r.

tpapp:
This fixes https://github.com/tpapp/julia-repl/issues/13 using the solution
already employed by `Base.Terminals`, see
https://github.com/JuliaLang/julia/commit/c9f28da19bfe061280fa91120c08ee851e1ff31e

I am not an expert on terminal control characters, but since it is used by Base
it must have seen a lot of testing so I assume it is correct. I tested on Linux,
using various terminals (also with tmux), and ansi-term inside Emacs, and
existing behavior is maintained.

(the neatest solution, of course, would be factoring out the control characters
in Base.Terminals in such a way that they could be sent to any `::IO`, this is but
a quick workaround).

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -293,7 +293,7 @@ end
 
 function move_cursor_up_while_clearing_lines(io, numlinesup)
     for _ in 1:numlinesup
-        print(io, "\u1b[1G\u1b[K\u1b[A")
+        print(io, "\r\u1b[K\u1b[A")
     end
 end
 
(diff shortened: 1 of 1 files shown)


timholy:
Thanks. I don't know much about this either (I learned as I went along). I will leave open for at least 24hrs for anyone to comment. One person who knows a lot about terminal characters is @Evizero, wondering if you might take a look?

Evizero:
I am afraid i don't know much about this one

tpapp:
I have now tested this PR with
```julia
using ProgressMeter
@showprogress for i=1:20
    sleep(0.1)
end
```
on all combinations of

1. `xfce4-terminal`, `xterm`, `gnome-terminal`, `ansi-term` in `Emacs 25.3`,
2. bare terminal, [`screen`](https://www.gnu.org/software/screen/), [`tmux`](https://github.com/tmux/tmux/wiki)

and it seems to work fine. Again, given that `Base.Terminals` is using these control characters, I hope it would be safe to merge.

timholy:
Thanks for tackling this!

PR #90, opened 2018-02-23
Use `stderr` instead of `STDERR` on newer Julia

martinholters:
Again, this depends on a not-yet-merged feature of Compat which I tentatively assume to be included in the next Compat version. (Tests pass locally when checking out the respective branch of Compat.)


martinholters:
Updated for the new approach taken by Compat. Should work once the Compat 0.55.0 tag has been merged.

Issue #86, opened 2017-12-08
Errors when progress is zero


Keno:
I get a regular callback with progress and sometimes that progress is zero causing an error:
```
julia> using ProgressMeter

julia> p = Progress(100, 1)
ProgressMeter.Progress(100, 1.0, 0, 1.512776740731284e9, 1.512776740731284e9, false, "Progress: ", 41, ProgressMeter.BarGlyphs('|', '█', '█', ' ', '|'), :green, Base.TTY(uRawFD(10) open, 0 bytes waiting), 0)

julia> update!(p, 0)
ERROR: InexactError()
Stacktrace:
 [1] trunc(::Type{Int64}, ::Float64) at ./float.jl:672
 [2] #updateProgress!#5(::Array{Any,1}, ::Symbol, ::Function, ::ProgressMeter.Progress) at /home/keno/.julia/v0.6/ProgressMeter/src/ProgressMeter.jl:156
 [3] #update!#9(::Array{Any,1}, ::Function, ::ProgressMeter.Progress, ::Int64) at /home/keno/.julia/v0.6/ProgressMeter/src/ProgressMeter.jl:231
 [4] update!(::ProgressMeter.Progress, ::Int64) at /home/keno/.julia/v0.6/ProgressMeter/src/ProgressMeter.jl:230
```

PR #91, opened 2018-03-09
Added ability to show 0 progress and reset progress bar.

ry-dgel:
As per #86, an error occurs when update!(p, 0) is run. This comes up due to an attempt to round an infinite eta time.

I've fixed this issue by adding a modified version of `durationstring()` that takes in a nullable, and will output "N/A" if the eta time is null. Coupled with this, `updateProgress!` now uses a nullable for `eta_sec` which will be set to null if `percentage_complete == 0`.

For a simulation that I'm working on, each step takes quite some time, so I was looking for a way to show the bar with 0% progress to at least indicate the the simulation had started. So, I added the `start!()` function which does exactly this. This function can also be used to restart a progress bar, resetting the timing and counter.

This is also my first ever pull-request and contribution to a public project, let me know how I did!

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -5,7 +5,7 @@ module ProgressMeter
 using Compat: stderr, lastindex, printstyled
 using Compat.Printf: @sprintf
 
-export Progress, ProgressThresh, BarGlyphs, next!, update!, cancel, finish!, @showprogress
+export Progress, ProgressThresh, BarGlyphs, start!, next!, update!, cancel, finish!, @showprogress
 
 """
 `ProgressMeter` contains a suite of utilities for displaying progress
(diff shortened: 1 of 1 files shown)


ry-dgel:
Travis is upset with my code not working on 0.7 builds. As I don't have 0.7 setup and running I can't immediately fix these issues in a cross compatible way, but for a 0.7 version it'll be simple to change to using Union{Some(T), Nothing} where T<:Number.

martinholters:
I'm not sure the introduction of a `Nullable` and the change to `durationstring` is needed here. Setting `eta = "N/A"` directly in `updateProgress!` should be ok. Alternatively, pull the definition of `msg` into the `if`/`else` and completely leave out the `ETA:` part if `iszero(percentage_complete)`. Personally I think I'd prefer the latter.

timholy:
Thanks for your contribution, @ry-dgel, and sorry I didn't react to your first public pull request! I've been really swamped the last few months.

Given the long latency, I can take it from here or coach you through it. Which would you prefer? I created the branch that become #98 before realizing this was here.

timholy:
I'm going to go ahead and merge #98. That doesn't fix the "show 0% progress" issue (tricky because when the computation is fast, you generally don't want to show any progress bar), so I'd be happy to entertain a PR focused on that issue.

PR #94, opened 2018-04-09
counterfirst

gustafsson:
I'd like to be able to specify where the progress meter should start so that I can continue a long-running process that is already half-finished.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -49,11 +49,12 @@ end
 
 """
 `prog = Progress(n; dt=0.1, desc="Progress: ", color=:green,
-output=stderr, barlen=tty_width(desc))` creates a progress meter for a
-task with `n` iterations or stages. Output will be generated at
-intervals at least `dt` seconds apart, and perhaps longer if each
-iteration takes longer than `dt`. `desc` is a description of
-the current task.
+output=stderr, barlen=tty_width(desc), resumefrom=0)` creates
+a progress meter for a task with `n` iterations or stages,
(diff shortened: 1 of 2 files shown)


martinholters:
Should be documented.

An alternative API could be something like `pause!(p)`/`resume!(p)`, so that the ETA would be based on the time spent before and after the pause, but not during. Not totally sure I like that better, though.

gustafsson:
Naming things is hard. I renamed `conterfirst` to `resumefrom` and added some documentation.

The idea with this PR I intend to cover the case when the process was killed and then resumed at a later point, so there wouldn't be a previously paused instance of `Progress` to call with `resume!`.

As a side effect I think this interface could also be used for use case of pause/resume. Though the user would have to deal with what they want their output to look like during the "pause".

martinholters:
I do like `resumefrom` better than `counterfirst`. And you're right, of course, for the killed process scenario, my proposal of `pause`/`resume` wouldn't cut it.

martinholters:
LGTM and I've resolved the conflicts, but would like to hear @timholy's opinion.

timholy:
Sorry for the absence here. This is a good idea.

As you say, naming is probably the hardest bit. Is there any reason not to call it `start`? For example, here's the julia 0.7 docstring for `range`:
```julia
  range(start; length, stop, step=1)
```
where either `length` or `stop` must be specified. Consistent with this, I wonder if we should call this keyword `start`?

timholy:
Needs to pass tests before it can be merged.

timholy:
Thanks @gustafsson!

PR #99, opened 2018-07-03
Fix depreciations in tests

adamslc:
Just a few quick and easy fixes to make the tests stop complaining.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,1 @@
-julia 0.6
-Compat 0.55.0
+julia 0.7-
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
+  - 0.7
   - nightly
 notifications:
   email: false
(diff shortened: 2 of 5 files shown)


timholy:
Another option is to adopt 0.7 wholesale. I'd be fine with either approach.

martinholters:
Dropping 0.6 support completely would also be fine with me. The changes here to keep compatibility with 0.6 appear to be so small, however, that it might be _less_ work than getting rid of Compat entirely. I leave it up to you, @adamslc.

adamslc:
OK, I've gone ahead and fully transitioned to 0.7.

timholy:
Thanks! Can you also remove the 0.6 line from `.travis.yml`? Then I think we can merge this.

REQUIRE
```diff
@@ -1,2 +1 @@
-julia 0.6
-Compat 0.55.0
+julia 0.7-
```
martinholters: Not that it matters much, but should we rather set this to `0.7.0-alpha`?

timholy:
Thanks, @adamslc (and @martinholters for his input here).

REQUIRE
timholy: Hmm, now that I think about it I guess this allows early dev versions too. Well, I don't really think anyone would be using anything older than the alpha anyway...

PR #100, opened 2018-07-11
RFC: Add Project.toml and set version to 1.0.0

martinholters:
The downside at present is that releases require manually bumping the version here in addition to tagging/creating a release on Github. As releases here don't occur too often, that should hopefully be OK. We might also immediately bump the version as part of this PR and and tag a release. Extrapolating from the API stability this package was able to maintain in the past, calling it 1.0.0 might be an option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/Manifest.toml
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ os:
 julia:
   - 0.7
   - 1.0
+  - 1.1
+  - 1.2
   - nightly
 notifications:
   email: false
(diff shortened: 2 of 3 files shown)


martinholters:
Ohhhhh, #100 - might be a sign that we indeed should use this PR to bump the version to 1.0.0 :smile:

timholy:
Love the numerology! Does this solve a problem now, or is it just for the future?

martinholters:
Just for the future:
```
┌ Warning: packages will need to have a [Julia]Project.toml file in the future
└ @ Pkg.Types Types.jl:582
```
Assuming that we're late enough in the Julia 0.7 release cycle that no more breaking changes occur, and that indeed no breaking changes occur in 1.0 that don't trigger a deprecation warning in 0.7, this should be enough to make ProgressMeter Julia 1.0 compatible :tada:

timholy:
I rebased and extended this. Thanks @martinholters!

PR #103, opened 2018-08-07
fixes for Julia 1.0

tanmaykm:
- used new iterator form
- used `seed!` instead of `srand`

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -380,19 +380,17 @@ struct ProgressWrapper{T}
 end
 
 Base.length(wrap::ProgressWrapper) = Base.length(wrap.obj)
-Base.start(wrap::ProgressWrapper) = (Base.start(wrap.obj), true)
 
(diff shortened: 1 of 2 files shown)


tanmaykm:
Yes, that should make it much more compact. Updating in a bit.

saschatimme:
Can we get this merged and cut a new release? I assume I am not the only one who depends on ProgressMeter :)

garrison:
I just ran the tests locally and all the progress meters look correct as far as I can tell.  @timholy, any other steps before cutting a new release?  I would suggest not merging #100 until after this release -- there is nothing (yet) that requires a Project.toml file, so I think it makes sense to wait until that becomes the canonical way of expressing project dependencies.  I myself have run into a few issues here and there with projects that include a Project.toml, and while I think most of the bugs have been ironed out, few other packages seem to have them to date so I see no reason to mess with it before getting a release out that works on julia 1.0.

I'll let Tim decide what the next version should be called :-).

PR #105, opened 2018-08-10
Drop support for julia 0.7 prereleases

garrison:
No urgent need for this, but why not? :slightly_smiling_face:


garrison:
I figured it made sense to do this before the next release, since I've seen METADATA.jl pull requests declined/delayed in the past for `REQUIRE`ing pre-releases of since released versions of julia.

PR #108, opened 2018-08-14
map and pmap to fix #32

### LINKED ISSUE #32: Working with pmap

zsunberg:
This PR will add support for both `map` and `pmap`

Primary usage will be through `@showprogress`, e.g.
```julia
@showprogress pmap(x->x^2, 1:10)
```

I have implemented the `progress_map` function, which uses channels to pass progress information back to the main process that prints the progress meter. It works for both `map` and `pmap`. Now I just need to modify `showprogress` to convert `map` and `pmap` calls into `progress_map` calls.

Comments welcome.


zsunberg:
Ok, I think this is nearly ready to merge.

There is one problem - currently ProgressMeter has to be loaded on all processes for this to work. I believe this is because the anonymous function defined on [line 575](https://github.com/zsunberg/ProgressMeter.jl/blob/8a7d885a82337c38d815d436e6e01ea18e5a42ff/src/ProgressMeter.jl#L575) apparently is defined only in the ProgressMeter module. I am not sure how to fix it.

I have put a fair amount of effort into creating concise code that works in all the cases I could think of, so I hope it can get merged soon :)

djsegal:
Is this like a cruel joke? I don't really care if it's perfect, but can we push something that's v0.7 compatible?

It seems like:

+ https://github.com/slundberg/PmapProgressMeter.jl

works perfectly in every place I've ever used it?

zsunberg:
Not sure what the cruel joke is - this PR is v0.7 compatible. Just need a maintainer to merge it @timholy . I think there might be another package that uses the builtin julia logging that people are switching to.

zsunberg:
@garrison are you the right one to ask to merge this?

timholy:
Sorry for the unconscionable delay, I have too many packages that I maintain and need to ditch responsibility for some. I'll give more people rights to this one.

timholy:
Turns out there were 5 people already authorized to merge PRs here, but I added you too @zsunberg.

Issue #52, opened 2016-07-30
Nested progress bars


juliohm:
Is it possible to produce nested progress bars as in a big task composed of smaller tasks?

grero:
I would also be interested in this ability.

juliohm:
Issue #42 is related, I am not sure about making it a plot, but I like the visual with multiple bars stacked as an initial implementation.

juliohm:
It seems that the feature was implemented by @IsaacTay in #113. Could you please to documentation? Do we have documentation in the README?

bhvieira:
Not sure about docs, just stumbled upon the project, but see the examples added with that PR @juliohm 

Here's (an adapted) one:
```
ProgressMeter.@showprogress "Level 0 " for i in 1:10
        ProgressMeter.@showprogress " Level 1 " 1 for i2 in 1:10
            ProgressMeter.@showprogress "  Level 2 " 2 for i3 in 1:10
                sleep(0.5)
            end
        end
end
```

juliohm:
Thank you @bhvieira , marking the issue as solved then. :+1:

PR #113, opened 2018-10-12
Added offset argument to allow for nested progress bars

IsaacTay:
Adding feature from #52

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -86,11 +86,11 @@ mutable struct Progress <: AbstractProgress
 end
 
 Progress(n::Integer, dt::Real, desc::AbstractString="Progress: ",
-         barlen::Integer=tty_width(desc), color::Symbol=:green, output::IO=stderr,
-         offset::Integer = 0) =
+         barlen::Integer=tty_width(desc), color::Symbol=:green, output::IO=stderr;
+         offset::Integer=0) =
     Progress(n, dt=dt, desc=desc, barlen=barlen, color=color, output=output, offset=offset)
 
-Progress(n::Integer, desc::AbstractString, offset::Integer = 0) = Progress(n, desc=desc, offset=offset)
+Progress(n::Integer, desc::AbstractString, offset::Integer=0) = Progress(n, desc=desc, offset=offset)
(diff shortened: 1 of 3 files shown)


src/ProgressMeter.jl
```diff
@@ -111,40 +114,52 @@ mutable struct ProgressThresh{T<:Real} <: AbstractProgress
...
+    p.offset = offset
     t = time()
     if p.counter >= p.n
         if p.counter == p.n && p.printed
             percentage_complete = 100.0 * p.counter / p.n
             bar = barstring(p.barlen, percentage_complete, barglyphs=p.barglyphs)
             dur = durationstring(t-p.tfirst)
             msg = @sprintf "%s%3u%%%s Time: %s" p.desc round(Int, percentage_complete) bar dur
+            print(p.output, "\n" ^ p.offset)
+            print(p.output, "\n" ^ p.numprintedvalues)
```
timholy: Combine these into `print(p.output, "\n"^(p.offset + p.numprintedvalues)` (it's approximately twice as fast, and easier to see the logic). This also applies several places below.

IsaacTay:
Do you think it would be a better idea to implement it as a wrapper to AbstractProgress? It would allow for more flexibility, but it does require the use of an IOBuffer, which would slow it down quite a bit.

IsaacTay:
[asciicast](https://asciinema.org/a/OopNqvmpHRLqpJyRsce2DSIiI)

timholy:
I really like this and am ready to merge this. (Let's not do the wrapper idea for now, given the performance issue.)

@juliohm and @grero, can you please test this and see if it fits your needs?

timholy:
CC also @ChrisRackaukas who gave the idea a thumbs-up.

ChrisRackauckas:
This looks fantastic.

juliohm:
Beautiful! :heart:

grero:
This is really cool and exactly what I was looking for. Thanks!

timholy:
@IsaacTay you have many fans, I hope to see you around the Julia community in the future!

Tagged release 0.8, when merged this will be broadly available.

PR #114, opened 2018-10-14
Fix dropping of color in printing of the first line in Atom

saschatimme:
I noticed that in Atom the first line of the progress bar wasn't printed with color (but additional values provided by `showvalues` are). With this also the first line is now printed with color.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -304,16 +304,12 @@ function move_cursor_up_while_clearing_lines(io, numlinesup)
 end
 
 function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
-    if isdefined(Main, :ESS) || isdefined(Main, :Atom)
-        print(io, "\r")
-        printstyled(io, s; color=color)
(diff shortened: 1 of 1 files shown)


src/ProgressMeter.jl
```diff
@@ -305,7 +305,8 @@ end
 
 function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
     if isdefined(Main, :ESS) || isdefined(Main, :Atom)
```
timholy: Could we refactor this to
```julia
function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
    print(io, '\r')
    printstyled(io, s; color=color)
    if isdefined(Main, :IJulia)
        Main.IJulia.stdio_bytes[] = 0
    elseif isdefined(Main, :ESS) || isdefined(Main, :Atom)
    else
        print(io, "\u1b[K")
    end
end
```

saschatimme:
Just refactored the function as you suggested.

timholy:
Thanks!

Issue #101, opened 2018-07-14
feature request: progress for unknown number of steps


tpapp:
Suppose I am reading an *ex ante* unknown number of records from a large file. It would be nice to do something like

```julia
prog = ProgressOnline("reading records")
while !eof(io)
    line = readline(io)
    ProgressMeter.update!(prog) # would just increment a counter, no second arg
    ... # process line
end
ProgressMeter.update!(prog, nothing) # done
```

while would display something like
```julia
reading records: 42000
```

ChrisRackauckas:
This would be solved by https://github.com/timholy/ProgressMeter.jl/issues/102 since that logging information lets you give the update percentage.

tpapp:
#102 is nice, but this issue is about the case where I **don't know** the percentage, since I don't know the total.

zsunberg:
Just out of curiosity, what is the advantage of this over just printing with a carriage return at the beginning?

```
i = 0
while !eof(io)
    line = readline(io)
    print("\rProcessed $i lines")
    ... # process line
end
println()
```

saschatimme:
@zsunberg  The implementation in #115  has a couple of advantages over your suggested solutions. 
It also automatically keeps track of the passed time and prints it nicely.
Also you can also easily specify additional values to be printed (with the `showvalues` keyword). Sure, you could all do this by your own, but it is also nice if everybody doesn't have to reinvent the wheel :)

Another major benefit of the implementation in #115 is that it updates the current information only every, let's say 0.1 seconds. Depending on the task this reduces the overhead of the progress bar immensely since IO operations are usually quite expensive.

zsunberg:
Great, thanks for the explanation

martinholters:
Closed by #115.

PR #115, opened 2018-10-14
Add a progress meter with unspecified number of iterations.

saschatimme:
This addresses #101 from @tpapp.
I am not super happy with the naming, so if somebody has a better idea :)

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -3,7 +3,7 @@ module ProgressMeter
 using Printf: @sprintf
 using Distributed
 
-export Progress, ProgressThresh, BarGlyphs, next!, update!, cancel, finish!, @showprogress, progress_map, progress_pmap
+export Progress, ProgressThresh, ProgressUnknown, BarGlyphs, next!, update!, cancel, finish!, @showprogress, progress_map, progress_pmap
 
 """
 `ProgressMeter` contains a suite of utilities for displaying progress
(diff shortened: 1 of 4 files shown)


timholy:
I like it. Naming is hard. Here are some alternatives I thought of:

- `ProgressIterations`
- `ProgressUnknown`
- `AreWeThereYet`
- `ProgressCounter`

timholy:
Maybe post a screen shot for people who don't want to build this to see what it looks like?

src/ProgressMeter.jl
```diff
@@ -203,6 +237,34 @@ function updateProgress!(p::ProgressThresh; showvalues = Any[], valuecolor = :bl
...
+        msg = @sprintf "%s %d \t Time: %s" p.desc p.counter dur
+        move_cursor_up_while_clearing_lines(p.output, p.numprintedvalues)
+        printover(p.output, msg, p.color)
+        printvalues!(p, showvalues; color = valuecolor)
+        # Compensate for any overhead of printing. This can be
+        # especially important if you're running over a slow network
+        # connection.
+        p.tlast = t + 2*(time()-t)
+        p.printed = true
+    end
```
timholy: Add `return`? Just for return-type stability.

saschatimme:
Here is how it looks: [asciicast](https://asciinema.org/a/SYwEAcPLzmfpA2xmUpeLOoTe4)

saschatimme:
> I like it. Naming is hard. Here are some alternatives I thought of:
> 
> * `ProgressIterations`
> * `ProgressUnknown`
> * `AreWeThereYet`
> * `ProgressCounter`

You mean that you like `ProgressOnline`?

src/ProgressMeter.jl
saschatimme: Good catch. Added it.

timholy:
I like the content. I'm not wild about `ProgressOnline`, but it's not obviously worse than the alternatives (though I confess a certain genuine fondness for `AreWeThereYet` :smile:). Anyone else want to chime in?

saschatimme:
I think I like `ProgressUnknown` the most by now. Do we just want to go with that?

saschatimme:
@timholy I would update the PR if you are happy(ish) with `ProgressUnknown` such that we can get this merged.

timholy:
Sure let's go with `ProgressUnknown`. You may not have seen this, but care to respond to https://github.com/timholy/ProgressMeter.jl/issues/101#issuecomment-430488777?

timholy:
Thank you!

timholy:
Just noticed #113. I will tag after taking a chance to review that, too. (Tomorrow, not today.)

PR #118, opened 2018-10-22
Make it possible to set the counter in `ProgressUnknown` by hand.

saschatimme:
This is a follow up to #115. I noticed that I missed two use cases
1) Sometimes you want to increase the counter by more than 1 or even decrease it again
2) Sometimes you want to update additional informations displayed by `showvalues` but not the counter

Both are fixed by adding the possibility to set the counter manually (by the same API as `Progress` already uses).

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -295,12 +295,12 @@ the current value.
 
 You may optionally change the color of the display. See also `next!`.
 """
-function update!(p::Progress, counter::Int; options...)
+function update!(p::Union{Progress, ProgressUnknown}, counter::Int; options...)
     p.counter = counter
     updateProgress!(p; options...)
 end
 
-function update!(p::Progress, counter::Int, color::Symbol; options...)
+function update!(p::Union{Progress, ProgressUnknown}, counter::Int, color::Symbol; options...)
     p.color = color
     update!(p, counter; options...)
 end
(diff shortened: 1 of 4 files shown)


src/ProgressMeter.jl
```diff
@@ -316,17 +316,13 @@ function update!(p::ProgressThresh, val, color::Symbol; options...)
     update!(p, val; options...)
 end
 
-function update!(p::ProgressUnknown; options...)
-    p.counter += 1
+function update!(p::ProgressUnknown; increasecounter=true, options...)
```
timholy: What's the `increasecounter` for?

saschatimme: The idea was that you could call `update!` without increasing the counter, e.g., if you only want to update the values display in `showvalues`:
```julia
update!(progress; increasecounter=false, showvalues=[(:myKey,  myValue)])
```

But I also see now that this is somewhat duplicate to the `update!(prog, counter)` method where you would just need to keep track of the counter manually.

saschatimme:
@timholy I removed the `increasecounter` keyword argument from the update function. Instead I noticed that it would be nicer that `ProgressUnknown` also supports the `next!` method for simply increasing the counter. If a user needs more control about the value of the counter `update!` should be used.

I also updated the section title for `ProgressThresh` slightly to better distinguish it from `ProgressUnknown`. 

It would be nice if this could be merged soonish. I would also appreciate if there would be a new release then since I depend on this feature for a new release of my package.

PR #120, opened 2018-12-03
allow one to transition between empty and full

KristofferC:
Enables slightly smoother progress bar which is nice when the width is not so big.

Link to asciinema below (they seem to have some bug in their monofont width rendering so it shifts a bit but this doesn't happen in the terminal):

https://asciinema.org/a/2BSeu9De6gO8UC05B0jDDx0jI

The first three bars are with a transition and the last one is with the current default.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -28,7 +28,7 @@ Holds the five characters that will be used to generate the progress bar.
 mutable struct BarGlyphs
     leftend::Char
     fill::Char
-    front::Char
+    front::Union{Vector{Char}, Char}
     empty::Char
     rightend::Char
 end
@@ -76,7 +76,7 @@ mutable struct Progress <: AbstractProgress
                       color::Symbol=:green,
                       output::IO=stderr,
                       barlen::Integer=tty_width(desc),
-                      barglyphs::BarGlyphs=BarGlyphs('|','█','█',' ','|'),
+                      barglyphs::BarGlyphs=BarGlyphs('|','█', ['▏','▎','▍','▌','▋','▊','▉'],' ','|',),
                       offset::Int=0)
         counter = 0
         tfirst = tlast = time()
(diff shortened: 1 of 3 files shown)


KristofferC:
I asked someone to test on windows and they said:

> The third and fourth style can be shown correctly, while the first and the second ones are just 口口口 (edited)

So perhaps we should only default to using the smoother one on Linux + Mac?

src/ProgressMeter.jl
```diff
@@ -28,7 +28,7 @@ Holds the five characters that will be used to generate the progress bar.
 mutable struct BarGlyphs
     leftend::Char
     fill::Char
-    front::Char
+    front::Union{Vector{Char}, Char}
```
zsunberg: Doesn't it make more sense to have this always be a `Vector{Char}` and just have a
```
BarGlyphs(le::Char, fi::Char, fr::Char, em::Char, re::Char) = BarGlyphs(le, fi, [fr], em, re)
```
constructor for backwards compatibility? That way, anyone using a BarGlyphs can always be sure that `front` is a vector.

KristofferC: It is possible but I would have to tweak some things slightly. Right now when the `front` is a vector and we want to draw, say, `2.05` bars, we will notice that for the `0.05` it is better to just print the `empty` than even the first character in the `front` vector.

However, for some bars is not correct (for example the `[=> ]` bar). Here we want to always print the `front` char because it is not a transition between `empty` and `full`.

The change I could make is to require someone to give the `front` vector with the `empty` and `fill` character also included. So it would be `[' ', '▁' ,'▂' ,'▃' ,'▄' ,'▅' ,'▆', '▇', '█']` and make sure that we always print one of the `front`chars. It is a bit redundant perhaps but maybe more general.

KristofferC: The reason I think we need the last one is in case we have something like `3.99` bars to show. Then for the default bar we want to show the full character `█` but if we have the restriction that we only show things in `front` then that is not available.

zsunberg: Oh, ok, I thought the procedure was to only display the `fill` character once that entire block has been completed (i.e. only display at 4.0). That makes the most sense to me, but it looks like the old way included "round" (in line 419) anyways. I am fine with either rounding or truncating though :)

KristofferC: Hm, I don't think we will display all the blocks for the same amount of time in that case.

It seems the current way that is implemented allows one to not have to repeat yourself at the cost of some type checking. In my view it is worth it to provide a simpler interface to the user.

zsunberg: Sure, sounds ok to me

zsunberg:
@KristofferC , were you still planning to make the default platform-specific for this? I don't see that in the code

KristofferC:
I committed but seems I forgot to push it.

zsunberg:
Awesome, thanks!

PR #121, opened 2018-12-13
Fix the elapsed time displaying

machakann:
Hi, please take a glance at the screenshot here.

ss

The displayed time in seconds has been rounded up to "60" if it was larger than 59.5. This pr fix this.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -444,7 +444,7 @@ function durationstring(nsec)
     hours = div(r,60*60)
     r = r - 60*60*hours
     minutes = div(r, 60)
-    seconds = r - 60*minutes
+    seconds = floor(r - 60*minutes)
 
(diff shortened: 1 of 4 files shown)


timholy:
Seems fine. Can you add a test? (You can call `durationstring` directly in the test.)

machakann:
I will try. I'm going to add a new file, `test/test_durationstring.jl`. Please let me know if there is another appropriate place to insert the test.

timholy:
Perhaps name it `core.jl` and run it first in `runtests.jl`. That way `core.jl` can grow to accomodate more "internal" tests over time without encouraging people to create separate files for each function in this package.

machakann:
Thank you, I have done.

machakann:
OK, I have done.

timholy:
Thanks!

PR #124, opened 2018-12-30
added alwaysflush option to fix #117

### LINKED ISSUE #117: Progress only displayed on exceptions and completed runs.
In Jupyter Lab:

peek 2018-10-21 18-06

Note that the progress bar only shows up once I manually interrupt the kernel. My counter, however, is printing just fine.

zsunberg:
By default now whenever this prints the progress meter, it will also flush the stream (see #117). Does everyone else think this is a good idea? Perhaps there are unforeseen consequences of flushing that I have not thought of.


timholy:
Seems fine to me. If the latency isn't a big deal then we could make it non-optional and get rid of one parameter.

zsunberg:
Yeah, I will just remove the option - it seems reasonable to do it every dt

PR #129, opened 2019-03-24
Use stdout's displaysize for size computation

tkluck:
Surprisingly, the argumentless default returns the values of the environment variables `ROWS` and `COLUMNS`, which (at least on gnome-terminal) are not necessarily up-to-date with window resizes.
Adding `stdout` gives a result that corresponds to current size.

This prevents the following situation:
Schermafdruk van 2019-03-24 10-22-49

A followup feature could be to poll this width to stay up-to-date with window resizes.


timholy:
Thanks!

PR #93, opened 2018-03-29
Fix Jupyter issue

xukai92:
I've seen this in Jupyter.

```
Training...  1%|                                                  |  ETA: 0:01:39
  epoch:     5
Training...  2%|█                                                 |  ETA: 0:01:34
  epoch:     11
Training...  3%|██                                                |  ETA: 0:01:31
  epoch:     17
Training...  5%|██                                                |  ETA: 0:01:29
  epoch:     23
Training...  6%|███                                               |  ETA: 0:01:28
  epoch:     29
Training...  7%|████                                              |  ETA: 0:01:27
  epoch:     35
Training...  8%|████                                              |  ETA: 0:01:25
  epoch:     41
  neg_elbo:  195.79621860107238
```

I used IJulia's function to clear the output to solve this issue. There is blink when using the function but I think it's better than having things above.

Issue #71, opened 2017-05-28
ProgressMeter.next! prints as many bars as updates (on Jupyter notebook)


davidbp:
Hello,

When I use ``ProgressMeter.next!(progressbar; showvalues = [(:epoch, epoch), (:acc, acc)])`

I get the following behaviour

```
Progress:   5%|██                                       |  ETA: 0:00:13
  epoch:  1
Progress:  10%|████                                     |  ETA: 0:00:11
  epoch:  2
Progress:  15%|██████                                   |  ETA: 0:00:09
  epoch:  3
Progress:  20%|████████                                 |  ETA: 0:00:08
  epoch:  4
```

I was expecting a single bar that changed, not multiple of them.
I have tried with ``@showprogress``and it works as expected. This is the output:

```
fit!(percep, X_train, y_train, 20, 0.001)
Progress: 100%|█████████████████████████████████████████| Time: 0:00:07
```

For reproducible code here there is the notebook 
https://github.com/davidbp/jml/blob/master/MulticlassPerceptron/benchmarks/Perceptron_Benchmark.ipynb

damiendr:
I have the same problem in IJulia and it seems to come from `showvalues`.

timholy:
What version are you two running? If it's 0.5.1, can you try `Pkg.pin("ProgressMeter", v"0.5.0")`?

timholy:
Oh, wait, this seems like a special case. Can you provide a complete minimum working example?

damiendr:
The problem occurs with both 0.5.0 and 0.5.1 — I have IJulia 1.6.2.

Here is a working example:
```julia
using ProgressMeter
N = 100
p = Progress(100)
for i in 1:100
    update!(p, i; showvalues = [(:test, 1.0)])
    sleep(0.01)
end
```

Here as a notebook: https://gist.github.com/damiendr/4d932dfe60d8de17fbbe9a2d215edcfc#file-progressmeter-ipynb
Github renders it just as it appears on my machine except that I have the text in colors.

juliohm:
I have the same issue on IJulia whenever I try to print some variable besides just showing the progress bar.

marius311:
Dug into this since I was also hoping to have this work. Looks like it boils down to different behavior between notebook / REPL for:
![image](https://user-images.githubusercontent.com/1510968/38392355-83f75dd2-3927-11e8-8350-65aa2964e533.png)
and
[...]

PR #131, opened 2019-06-26
Allow to opt-in into clearing IJulia cell

saschatimme:
This add the possibility to clear output of IJulia cells to get a behaviour similar to the REPL. 
The difference to #93 is that this behaviour is completely optional **and** you have to opt-in into this behaviour. I think we should let make the end user the call whether he wants this behaviour or not while providing by default the safe choice.

Here is a video of the old and the new behaviour:
https://www.youtube.com/watch?v=nO8N1Cda0FA


zsunberg:
This seems ok to me - it is only needed when you use stdout though, right? I think the documentation should reflect this.

zsunberg:
Oh, I see - it is only in the presence of showvalues. Why does it work correctly when there are no showvalues? Because it only has to delete one line?

Is it possible to automatically detect whether the user has printed something else? If that is possible, we could include this behavior automatically if nothing else is printed.

findmyway:
Is it possible to have this feature merged?

timholy:
> Is it possible to have this feature merged?

Yes. But it would be great to have @saschatimme respond to @zsunberg's questions. I don't think we have to get the automatic part implemented in this PR, though.

arnavs:
Thanks for this @saschatimme; ended up using this in my own work just now.

xukai92:
> Why does it work correctly when there are no showvalues?

There was some explanations here https://github.com/timholy/ProgressMeter.jl/issues/71#issuecomment-379222272

> Is it possible to automatically detect whether the user has printed something else? If that is possible, we could include this behavior automatically if nothing else is printed.

There is. Basically we only need to trigger it when `showvalues` is not empty for IJulia. But I think the idea of this PR is to make this IJulia clear output behaviour explict rather than automatic.

I'm also looking forward to have this feature merged :)

xukai92:
As a note, one way to automate this is by the following
```julia
function ProgressMeter.move_cursor_up_while_clearing_lines(io::Main.IJulia.IJuliaStdio, numlinesup::Int)
    if numlinesup > 0 && isdefined(Main, :IJulia) && Main.IJulia.inited
        Main.IJulia.clear_output(true)
    end
end
```
We can load this one when IJulia is defined using Requires.jl

xukai92:
Also, the current way to make it optional is by set the behaviour when creating the progress bar and passing the keyword agurment through all the functions. Wouldn't it be better to deal with this option in a package level? We can do something like below

```julia
const is_clearijulia = Ref(false) # off by default
clearijulia(flag=false) = (is_clearijulia[] = flag)
```

Users can create the progree bar as usual and turn this on and off by `clearijulia(true)` or `clearijulia(false)` globally.

PR #134, opened 2019-09-13
make @showprogress work with @distributed

rjkat:
I find that I often want to use `@distributed` loops with a reducer to compute a result. The code in the "tips for parallel programming" section of the README works well so thought I'd make a macro out of it. If there is a reason for not wanting to put the code from the README in the package then feel free to ignore/close this request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,23 @@ The first incantation will use a minimum update interval of 1 second, and show t
 
 The `@showprogress` macro wraps a `for` loop, comprehension, or map/pmap as long as the object being iterated over implements the `length` method and will handle `continue` correctly.
 
+The `@showprogressdistributed` macro is the equivalent of `@showprogress` and wraps `@distributed` for loops with a reducer.
+
+```julia
+using Distributed
+using ProgressMeter
+
+result = @showprogressdistributed (+) for i in 1:10
+    sleep(0.1)
+    i^2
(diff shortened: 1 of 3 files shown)


timholy:
This is quite nice. I'm not opposed to making this easier. Worth asking, though, whether there's a more elementary common core. For example, the `Channel`-based communication and progress bar updating could clearly generalize well beyond reduction operations.

zsunberg:
Yes, perhaps the most idealistic thing to do might be to have a ProgressMeter launch its own separate task by default and have `next` and `update!` communicate to it through channels. That would make all progress meters safe for multi-processing (and multi-threading?).

Anyways, on this particular PR, it seems better to me if the syntax was
```julia
@showprogress @distributed for ...
```

rjkat:
Both good suggestions. I've implemented the
```julia
@showprogress @distributed for
```
for `@distributed` for loops with and without reducers. Abstracting out the `Channel`-based progress updating seems like a worthy goal, but probably requires a bit of design thinking. Would you be happy to leave this for another time?

zsunberg:
Looks good to me! I'll give @timholy or anyone else a chance to comment and merge this tomorrow if nothing else comes up.

timholy:
Really nice work, @rjkat, thanks for the contribution!

PR #137, opened 2019-10-18
Avoid allocations due to kwarg default showvalues = Any[]

pablosanjose:
I was surprised that a function like `foo() = @showprogress for i in 1:10^7; rand(); end` produced 10^7 allocations. The reason, it seems, is the `showvalues = Any[]` kwarg default. I'm not sure why it is written like this, but making it an empty tuple removes one allocation per iteration, and moreover cuts the runtime overhead dramatically

Master
```
julia> @time foo()
Progress: 100%|███████████████████████████████████████| Time: 0:00:00
  0.831637 seconds (10.00 M allocations: 762.963 MiB, 14.18% gc time)
```

PR:
```
julia> @time foo()
Progress: 100%|███████████████████████████████████████| Time: 0:00:00
  0.411675 seconds (232 allocations: 13.703 KiB)
```

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -174,7 +174,7 @@ ProgressUnknown(desc::AbstractString) = ProgressUnknown(desc=desc)
 tty_width(desc) = max(0, displaysize(stdout)[2] - (length(desc) + 29))
 
 # update progress display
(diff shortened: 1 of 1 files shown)


timholy:
Thanks @pablosanjose, sorry I didn't notice this earlier!

PR #138, opened 2019-10-23
fix #136 by adding yield in progress_map

### LINKED ISSUE #136: visual feedback for `map` depends on presence of `sleep`
I observed that in some cases I would get visual feedback only at the very end or after more than 1 minute time when looping with `map`:

```
using ProgressMeter

@showprogress map(1:1000) do i
    A = rand(10000,1000)
    sum(A)
end;
```
This loop runs for about 20 seconds and only shows the progress bar at the very end. 

However, when I dress the code with a `sleep` everything is fine
```
@showprogress map(1:1000) do i
    sleep(0.000001)
    A = rand(10000,1000)
    sum(A)
end;
```
and I get continuous feedback on the progress. 

This happens in the REPL, in Juno, and in JupyterLab. 

Note: A similar construction with a `for` loop works fine with and without the `sleep`.

Screen Shot 2019-10-16 at 3 33 28 PM

zsunberg:
I think this is uncontroversial enough to just merge without review once the tests pass. Happy to hear if anyone has any feedback though.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,11 +1,11 @@
 import ProgressMeter
 using Test
 
-include("core.jl")
-include("test.jl")
-include("test_showvalues.jl")
+# include("core.jl")
+# include("test.jl")
+# include("test_showvalues.jl")
 include("test_map.jl")
-include("test_float.jl")
+# include("test_float.jl")
 
 println("")
 println("All tests complete")
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -768,6 +768,7 @@ function progress_map(args...; mapfun=map,
             vals = mapfun(other_args...; kwargs...) do x...
                 val = f(x...)
                 put!(channel, true)
+                yield()
                 return val
             end
             put!(channel, false)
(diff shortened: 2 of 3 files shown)


test/runtests.jl
```diff
@@ -1,11 +1,11 @@
...
 
-include("core.jl")
-include("test.jl")
-include("test_showvalues.jl")
+# include("core.jl")
+# include("test.jl")
+# include("test_showvalues.jl")
 include("test_map.jl")
-include("test_float.jl")
+# include("test_float.jl")
```
martinholters: Committed these by accident?

zsunberg: Yes! :man_facepalming: sorry! Is there a way to only run certain test sets so that I don't have to comment out ones to skip?

KristofferC: So what's the plan to fix it? :)

timholy: @zsunberg, if you want to just run that one test you can just navigate to the test directory and say `include("test_map.jl")`.

Ideally, this should have been fixed when it was first pointed out. Fixed in 237365084b3fd04bd1c8446d5dd225ebf6be86df

zsunberg: Yes, I apologize! I am moving to a new state this week, so it slipped through the cracks.

Issue #135, opened 2019-09-30
@showprogress for Threads.@threads


ianshmean:
Would it be possible? It's probably asking a lot!
Something functionally along the lines of the example at https://github.com/timholy/ProgressMeter.jl#tips-for-parallel-programming

Perhaps `@showprogress` should also handle the `wait(t)`. Something simple like this, perhaps:

```julia
x = Array{Float64}(undef, 100)
@showprogress "Completing parallel work.." Threads.@threads for i in 1:100
    x[i] = rand(Float64)
    sleep(0.1)
end
```

Wieke:
Using a simple spinlock seems to work, it ensures that only a single thread can call next at the same time.

```julia
    p = Progress(100)
    l = Threads.SpinLock()


    Threads.@threads for n = 1:100
        sleep(0.1)
        lock(l)
        next!(p)
        unlock(l)
    end
```

ianshmean:
Nice! I'm certain that this test has shortcomings, but it doesn't seem to slow things down in this case:
```julia
using ProgressMeter
function foo(n)
    x = Array{Array{Float64}}(undef,n)
    a = rand(1000,1000)
    Threads.@threads for i = 1:n
        x[i] = a .* rand(1)
    end
    return x
end
function fooprogress(n)
    p = Progress(n)
    l = Threads.SpinLock()
    x = Array{Array{Float64}}(undef,n)
    a = rand(1000,1000)
    Threads.@threads for i = 1:n
        x[i] = a .* rand(1)
        lock(l)
        next!(p)
        unlock(l)
    end
    finish!(p)
    return x
end

using BenchmarkTools
@btime fooprogress(100) # 105.004 ms (535 allocations: 770.61 MiB)
@btime foo(100) # 103.154 ms (340 allocations: 770.59 MiB)
```

ianshmean:
Or just the original `sleep(0.1)` example:
```julia
using ProgressMeter
function foo(n)
    Threads.@threads for i = 1:n
        sleep(0.1)
    end
end
function fooprogress(n)
    p = Progress(n)
    l = Threads.SpinLock()
    Threads.@threads for i = 1:n
        sleep(0.1)
        lock(l)
        next!(p)
        unlock(l)
    end
    finish!(p)
end

using BenchmarkTools
@btime fooprogress(100) # 2.043 s (1765 allocations: 117.45 KiB)
@btime foo(100) # 2.045 s (741 allocations: 20.47 KiB)
```

[...]

PR #139, opened 2019-10-29
Add ProgressThreads for Threads.@threads etc.

ianshmean:
Edit: Was `ProgressParallel`, now `ProgressThreads`

------
Based on @wieke's suggestion in #135 

This uses `Threads.SpinLock()` internally to ensure that only a single thread can increment the progress bar (update the counter and call `updateProgress!(p; options...)`) at the same time

```julia
using ProgressMeter
n = 100
p = ProgressThreads(n)
Threads.@threads for i = 1:n
    sleep(0.1)
    next!(p)
end
```

Please feel free to edit

N.B. I noticed that most of the test scripts are commented out on master. I just wanted to check that was intended.


timholy:
> N.B. I noticed that most of the test scripts are commented out on master. I just wanted to check that was intended.

Thanks, @ianshmean! Fixed in 237365084b3fd04bd1c8446d5dd225ebf6be86df. I caused a merge conflict for you, but fixed it as well. Let's see if everything passes.

ianshmean:
Travis was going slow there for MacOS..but 🎉

I did wonder about  making`@showprogress` work, but perhaps that's a future PR

README.md
```diff
@@ -194,9 +194,27 @@ for iter = 1:10
 end
 ```
 
+### Parallel processes
+
+For processes that are parallelized with `Threads.@spawn` or `Threads.@threads`,
```
KristofferC: What would an example with `Threads.@spawn` look like?

ianshmean: just testing.. getting some odd behavior on a basic loop on 1.3-rc4.. `@spawn` is only using 1 thread, even though `nthreads()==5`
```
@show Threads.nthreads() # =5
t = Threads.@spawn for i = 1:100
    println("$i on $(Threads.threadid())")
    sleep(0.1)
end
```

ianshmean: Just realized I'm misusing `@spawn`. I'm working on an example

ianshmean: @KristofferC that's proving to be a good question.. I can't figure out an example, so I'll remove mention of `@spawn`

ianshmean:
In response to @KristofferC's review question, this example uses `Threads.@spawn`. I don't know whether it's a decent example of using `@spawn` though..

```julia
using ProgressMeter
n = 100

threads = Threads.nthreads() - 1
tasks = Vector{Task}(undef, threads)

p = ProgressParallel(length(tasks) * n)
for t in 1:length(tasks)
    tasks[t] = Threads.@spawn for i in 1:n
        sleep(rand()/2)
        next!(p)
    end
end

wait.(tasks)
```

ianshmean:
Good point. I'll leave some ideas below for votes:

ianshmean:
`ProgressThreads(n)`

ianshmean:
`ProgressThreaded(n)`

ianshmean:
`ProgressMultithreaded(n)`

ianshmean:
`ProgressParallel(n)` (original)

ianshmean:
I've gone with the unanimous winner 👍🏻

timholy:
Very nice. Are we remotely worried about a failure while the lock is held?
```julia
lock(p.spinlocker) do
    p.color = color
    next!(p; options...)
end
```
will ensure it gets unlocked even if the body of the `do` loop throws an error. But I can't honestly think of a case where the current implementation should be a problem.

ianshmean:
Seems like the right thing to do, even so. I’ll switch it out.

ianshmean:
I've switched that out, improved the `@threads` test, and added a `@spawn` test. 

It turns out that `@threads` doesn't play nicely with `BitArray`s as they're not threadsafe [(issue)](https://github.com/JuliaLang/julia/issues/33750), so I've switched the examples to a float array.

yuyichao:
I don't know what you are doing so I don't know what fits your need the best. Just FYI that only bitarray is the problem,  `Array` of `Bool` is fine.

ianshmean:
Ok, good to know, thanks. I think these tests are ok being float arrays

ianshmean:
@timholy Agreed 👍🏻

KristofferC:
Is there any reason to not do this unconditionally? In other words, is the locking expensive enough that a separate type haS to be introduced?

ianshmean:
Good question. I can PR that approach to see how it tests.

Also, I just noticed that `next!(p::ProgressThreads, color::Symbol; options...)` is untested and will hang as it tries to spinlock within a spinlock. 

We should hold off merging this

ianshmean:
Seems like it works! I don't notice a performance regression, but only tested by eye:
See https://github.com/timholy/ProgressMeter.jl/pull/140

ianshmean:
Just to confirm, seems like repeatedly locking/unlocking isn't expensive (both results seem equivalent/in the noise):

```julia
function foo(n)
    l = Threads.SpinLock()
    for _ = 1:n
        lock(l) do
            sleep(0.00001)
        end
    end
end

function bar(n)
    for _ = 1:n
        sleep(0.00001)
    end
end

using BenchmarkTools
@btime foo(100) # 118.082 ms (706 allocations: 16.97 KiB)
@btime bar(100) # 118.373 ms (704 allocations: 16.94 KiB)
```

ianshmean:
Did some more tests over in #140. I much prefer that PR

(this PR needs a fix first if we go with it)

timholy:
Makes sense. Taking a lock is very cheap compared to manipulating the terminal. Great suggestion, @KristofferC, and thanks @ianshmean for implementing it!

ianshmean:
#140 won over this

PR #140, opened 2019-11-03
Add spinlocking for core methods, to support threading (alt to #139)

ianshmean:
Trying out @KristofferC's suggestion in #139 of just spinlocking by default.

All tests pass locally for me!

```julia
n = 100
p = ProgressMeter.Progress(n)
Threads.@threads for i = 1:(n)
    sleep(0.1)
    ProgressMeter.next!(p)
end
```

I didn't figure out getting `@showprogress` working, but I assume it would follow something like intercepting `@threads` and passing to a custom `showprogressthreads`?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,14 @@ env:
   matrix:
     - JULIA_NUM_THREADS=1
     - JULIA_NUM_THREADS=3 #3 allows the @spawn test to run effectively
+jobs:
+  exclude:
+  - julia: 0.7 # Disable Threads testing
+    env: JULIA_NUM_THREADS=3
+  - julia: 1.0 # Disable Threads testing
+    env: JULIA_NUM_THREADS=3
+  - julia: 1.1 # Disable Threads testing
+    env: JULIA_NUM_THREADS=3
 # script:
 #   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
 #   - julia --project -e 'using Pkg; Pkg.build();'
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ julia:
   - nightly
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e 'if VERSION < v"0.7-"; Pkg.clone(pwd()); else; import Pkg; Pkg.add(Pkg.PackageSpec(path=pwd())); end; Pkg.build("ProgressMeter")'
-  - JULIA_NUM_THREADS=2 julia -e 'Pkg.test("ProgressMeter"; coverage=true)'
+  - julia --project -e 'using Pkg; Pkg.build();
+  - JULIA_NUM_THREADS=2 julia --project -e 'Pkg.test(; coverage=true)'
 notifications:
   email: false
 git:
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ julia:
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia --project -e 'using Pkg; Pkg.build();
-  - JULIA_NUM_THREADS=2 julia --project -e 'Pkg.test(; coverage=true)'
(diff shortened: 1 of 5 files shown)


ianshmean:
More tests. Seems ok on performance.

Master (no spinlocker):
```julia
using BenchmarkTools
@btime foo(5000) # 6.244 s (35208 allocations: 849.50 KiB)
@btime fooprogress(5000) # 6.164 s (42883 allocations: 1.42 MiB)
```

This PR (spinlock everything):
```julia
using BenchmarkTools
@btime foo(5000) # 6.043 s (35991 allocations: 897.31 KiB)
@btime fooprogress(5000) # 5.605 s (47627 allocations: 1.47 MiB)

# @threads
@btime bar(5000) # 833.957 ms (35252 allocations: 853.72 KiB)
@btime barprogress(5000) # 833.926 ms (45655 allocations: 1.33 MiB)
```

```julia
using ProgressMeter
function foo(n)
    for _ = 1:n
        sleep(0.00001)
    end
end
function fooprogress(n)
    p = Progress(n)
    for _ = 1:n
        sleep(0.00001)
        next!(p)
    end
end

function bar(n)
    Threads.@threads for _ = 1:n
        sleep(0.00001)
    end
end
function barprogress(n)
    p = Progress(n)
    Threads.@threads for _ = 1:n
        sleep(0.00001)
        next!(p)
    end
end
```

timholy:
Awesome. I hate making more requests, but can you insert this into the `.travis.yml`?

```sh
script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
  - julia -e 'if VERSION < v"0.7-"; Pkg.clone(pwd()); else; import Pkg; Pkg.add(Pkg.PackageSpec(path=pwd())); end; Pkg.build("ProgressMeter")'
  - JULIA_NUM_THREADS=2 julia -e 'Pkg.test("ProgressMeter"; coverage=true)'
```

(Untested, but hopefully it will work.)

KristofferC:
I think you could also make a `matrix` with env variables and the test script is run on each entry (as in https://github.com/JuliaInterop/ZMQBuilder/blob/b1fdcfbde2dceda1af78205bf8a3d81fe4fa6a08/.travis.yml#L19)

ianshmean:
Looks like there's thread print issues <1.2 (and a problem on MacOS nightly..)

@KristofferC I didn't quite understand what you meant in your last suggestion

KristofferC:
Just suggested an easier way to run the tests with one and many threads than to change it in the test script.

ianshmean:
@KristofferC Got it. Nice and tidy. thanks

Looking at test results, is a compat of `julia = 1.2` acceptable for this package? The PR doesn't break previous usage, just when multi-threading, which seems like a lower-level julia issue

timholy:
Sorry for the long delay, I have been focusing on closing a bunch of Revise bugs and limitations.

Testing at #142 whether we can make the threading tests conditional. I am basing it off https://github.com/travis-ci/travis-ci/issues/7149#issuecomment-335825605, though I don't pretend to really understand Travis' yml format.

ianshmean:
@timholy I just tweaked travis on this PR to exclude specific jobs from the results of the build matrix, based on https://docs.travis-ci.com/user/build-matrix/#excluding-jobs

Seems like it worked!

ianshmean:
Oh, and while travis doesn't look like it triggered on the github checks box, it actually did. Must be a github bug?
https://travis-ci.org/timholy/ProgressMeter.jl/builds/610452866?utm_medium=notification&utm_source=github_status

ianshmean:
I can't see why the linux 1.2 failed.. I couldn't see a source of the error. That might just need a restart

timholy:
Thanks so much @ianshmean! I will work on a release tomorrow morning!

karajan9:
> ```julia
> sleep(0.00001)
> ```

Since I stumbled over this recently: `sleep` only goes down to `sleep(0.001)` as you can also see in the resulting time. But I think this shouldn't affect the results of the test.

PR #141, opened 2019-11-07
[RFC] Always print something before the first iteration.

fredrikekre:
I sometimes find it useful to get some feedback that the iteration has started. In cases where each iteration takes a while, you don't get any feedback until that iteration is done.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -197,7 +197,7 @@ function updateProgress!(p::Progress; showvalues = Any[], valuecolor = :blue, of
         return nothing
     end
 
-    if t > p.tlast+p.dt
+    if t > p.tlast+p.dt || p.counter == 0 # Always print the first time
         percentage_complete = 100.0 * p.counter / p.n
         bar = barstring(p.barlen, percentage_complete, barglyphs=p.barglyphs)
         elapsed_time = t - p.tfirst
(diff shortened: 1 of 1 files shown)


timholy:
This is a tough call. In cases where it finishes really quickly, I'd rather it didn't display anything at all.

I wonder if more heavily leveraging threading is one way to go here; if one thread just updates the progress meter, then we could get the best of all worlds.

fredrikekre:
> This is a tough call. In cases where it finishes really quickly, I'd rather it didn't display anything at all.

Hmm, I thought the "done" printing was always done anyway, and then this PR would at wors result in just another print.

timholy:
It's definitely aimed at long-running tasks, but also aims to be light weight. For tasks that take very little time the printing would dominate the run time.

I'm definitely torn. Let's see what other people think.

ianshmean:
How about an optional `showzero` arg that's `false` by default?

zsunberg:
Yeah, I like adding a keyword argument as @ianshmean  suggested, but I think it should always show by default. Obvious consistent behavior is more important than performance in things that print.

timholy:
Sounds like consensus is to print. Any objections if we merge this as-is and I or someone else adds the keyword argument?

fredrikekre:
TBH after learning that no printing is done at all (not even final printing) for very fast loops I am not sure I like this PR myself 🤷‍♂

PR #142, opened 2019-11-11
Try making threading conditional

timholy:
I'm playing with #140 to see if we can skip running with multiple threads on Travis for problematic Julia versions. CC @ianshmean


timholy:
Weird that travis didn't start. Anyone understand why?

timholy:
@ianshmean fixed this in #140

PR #147, opened 2020-02-16
Clear IJulia cell with a package level behaviour control

xukai92:
This is PR deals with the Jupyter notebook clear issues as discussed in https://github.com/timholy/ProgressMeter.jl/issues/71. Different from https://github.com/timholy/ProgressMeter.jl/pull/131, the behvaviour of clear IJulia output is controlled by a package level flag instead of passing a kerword argument to `Progress`:

```julia
ProgressMeter.clearijulia(true)
# or
ProgressMeter.clearijulia(false)
```

~~Currently the behaviour is disabled by default. Maybe we should enbale it by default?~~


xukai92:
@zsunberg @timholy

xukai92:
I make it enabled by default now. I think the best way to handle this if we do want this package level option over the keyword agurment option is to throw a warning for the first time this behaviour is triggered so users know what's happening.

zsunberg:
This seems good to me. I think enabled is the correct default. Glad that we could avoid the dependency on IJulia.

xukai92:
> I think the best way to handle this if we do want this package level option over the keyword agurment option is to throw a warning for the first time this behaviour is triggered so users know what's happening

Do we want this feature in this PR?

zsunberg:
I don't think a warning's necessary. What would it say? (Maybe I'm misunderstanding what this PR does)

xukai92:
> I don't think a warning's necessary. What would it say? (Maybe I'm misunderstanding what this PR does)

So so the way this PR (and other PRs on solving the Jupyter issue) is to use `IJulia.clear_output(true)` to clear the whole output of the cell. This means if there are other outputs, for example, a plot got printed during the loop of the progress bar, those outputs would be cleaned up as well.

In this case, we could throw a warning if user is in IJulia mode tell them that this would happen by default and they can disable it if they want.

zsunberg:
I see. I vote yes on some form of warning

We could also have three options:
- `ProgressMeter.ijulia_behavior(:warn)` which warns and clears (this would be default)
- `ProgressMeter.ijulia_behavior(:clear)' which just clears without a warning
- `ProgressMeter.ijulia_behavior(:append)` which does not clear

I am ok with the above or just warning once.

xukai92:
I implemented your "three options" proposal for this functionality in https://github.com/timholy/ProgressMeter.jl/pull/147/commits/36513a35ab23c9394856f6a7ca7c4efa003028bb.

xukai92:
@zsunberg Does the update look good to you?

zsunberg:
@xukai92 thank you! sorry this slipped through the cracks for me.

Yes, it looks good. I guess since timholy already approved, I will go ahead and merge.

xukai92:
Thanks a lot for the help!

timholy:
I've been crazy busy the last two weeks (Zoom is my life...), but things are easing up. How do people who've tried this like the `:warn` behavior? Is this something we're going to have for a while, or will it go away at some point?

timholy:
Also, should this be called v2.0?

DilumAluthge:
If this PR changed the default behavior, then I think it's probably a breaking change.

zsunberg:
Ah, I didn't think about version numbering when I merged this. My *feeling* is that this is a small change that should not be version 2.0. I would consider this a bug fix rather than a default behavior change, so maybe that is a reason to keep it as 1.x.

If the current default means we need to go to 2.0, I would advocate instead sticking to 1.x and changing the default to something that does the old behavior (`:append`), but also prints info to suggest using (`:clear`). Maybe call it (`:suggest`)? What do people think.

timholy:
Do we keep the "suggest" default forever or is this transitional? If it's transitional, does the warning effectively force people to change their code anyway? (Will the warning popping up in the middle of a lecture be too annoying?) If it's temporary what does it change to in the long run? Does it go back to `:append` or switch to `:clear`?

zsunberg:
Good point - you clearly have more experience with the long-term ramifications. I think it would be fine to tag 1.3 and call this a fix to incorrect behavior (it only affects the case where you are in juypter *and* printing values, and if you were doing that, it would have been very annoying anyways) or tagging it as version 2.0.

DilumAluthge:
If we call this PR a "fix to incorrect behavior", then it is consistent with SemVer to make the next release `1.3`.

timholy:
OK, to me that suggests we should also change the default to `const IJULIABEHAVIOR = Ref(IJuliaClear)`. Sound reasonable?

xukai92:
Although we would not expect one will take the behaviour off, the warning message does tells the user that they won't expect things like figures to be printed along the progress bar. Thus I think this message is still informative.

ianshmean:
Is this close to being resolved? It would be good to release a new version given a rather nasty deadlock issue that seems more common on 1.4.1 has been fixed on master

ianshmean:
The deadlock issue in the last release is rather frustrating. Can I release 1.3? Hit that thumbs up if you approve

cc. @timholy @zsunberg @DilumAluthge @xukai92 
x-ref #150

ianshmean:
Ok, I'll release 1.3 tomorrow eve

timholy:
xref #162

PR #148, opened 2020-03-06
add `start` keyword to resume progress from a higher iteration number

jgidi:
Hello!

I wanted the option to (re)start a ProgressMeter from some iteration number higher than 0, so I did some pretty straightforward changes, and then I found out that @gustafsson had a [pull request](https://github.com/timholy/ProgressMeter.jl/pull/94) addressing just this issue.

However, this pull request has been open since 2018, it had a few merge conflicts and also @timholy had suggested the name `start` for the keyword (which I agree) instead of `counterfirst` as proposed by the original author of the request.

Considering the former, I make this pull request over the present master branch to revive the idea and to make it faster/easier for the maintainers to adopt the changes, which can be automatically merged if they agree.

Finally, it seems important to disclaim that even when changes I make here are straightforward, they are very much the same ones proposed by @gustaffson (besides naming), and I do not want, by any means, to steal his work. Any thanks are to him.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -49,8 +49,8 @@ end
 
 """
 `prog = Progress(n; dt=0.1, desc="Progress: ", color=:green,
-output=stderr, barlen=tty_width(desc))` creates a progress meter for a
-task with `n` iterations or stages. Output will be generated at
+output=stderr, barlen=tty_width(desc), start=0)` creates a progress meter for a
+task with `n` iterations or stages starting from `start`. Output will be generated at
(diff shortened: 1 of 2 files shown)


gustafsson:
Yep, `start` is a better name. I rebased #94 on the most recent master and included your commit.

timholy:
Thanks @jgidi for restarting this. This version passes tests and #94 does not, but let's give @gustafsson another chance to get the tests passing. If that doesn't happen in the next few weeks, ping me again here and I'll merge this.

ianshmean:
#94 was merged, so I'm assuming this can be closed

PR #153, opened 2020-04-26
Switch from spinlock to reentrantlock

### LINKED ISSUE #152: Example on Threads in README results in deadlocks
The following example from **README** results in hanging for `n>=70` using 4 Threads
```using ProgressMeter
n = 10
p = Progress(n)
tasks = Vector{Task}(undef, n)
for i in 1:n
    tasks[i] = Threads.@spawn begin
        sleep(2*rand())
        next!(p)
    end
end
wait.(tasks)
```

ianshmean:
An attempt to fix #152

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -57,7 +57,7 @@ the current task.
 """
 mutable struct Progress <: AbstractProgress
     n::Int
-    spinlocker::Threads.SpinLock
+    reentrantlocker::Threads.ReentrantLock
     dt::Float64
     counter::Int
     tfirst::Float64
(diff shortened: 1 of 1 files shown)


ianshmean:
Going to merge as the docs confirm reentrantlock should be used
https://docs.julialang.org/en/v1/base/multi-threading/#Base.Threads.SpinLock
> SpinLock() 
> Create a non-reentrant, test-and-test-and-set spin lock. Recursive use will result in a deadlock. This kind of lock should only be used around code that takes little time to execute and does not block (e.g. perform I/O). In general, ReentrantLock should be used instead.

PR #154, opened 2020-04-27
update ci to 1.0, 1.4, nightly

ianshmean:


.travis.yml
```diff
@@ -3,23 +3,17 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
-  - 1.1
-  - 1.2
+  - 1.4
```
timholy, line 7: This could just be `1` so that the latest stable release gets run.

.travis.yml
```diff
@@ -3,23 +3,17 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
```
timholy, line 6: Removing testing for 0.7 requires that we remove support for 0.7. If I understand the automatic merge guidelines, that constitutes a breaking release so we'd need to go to 2.0.0. But I'm not certain.

.travis.yml
ianshmean: Nice! 👍🏻

.travis.yml
ianshmean: Ok, I didn't mean to be breaking. I thought 0.7 would be ok to remove given it was so temporary. But I like robust logic, so I'll add it back for now

PR #155, opened 2020-05-10
Dynamic terminal width

pengwyn:
Allows for a dynamic terminal width, fixing issues with line wrapping and allows for truncating of a showvalues line.

Does 3 related things:

1. Calculates `barlen` during the run, so that resizing the terminal is noticed.
2. Does not assume a showvalues line is only 1 line long.
3. Adds in a `truncate_lines` option for `update!` to force a showvalues line to be only 1 line long. This gives something like:

```
Progress: 100%|███████████████████████████████████████| Time: 0:00:20
  message:   0123∀ x ∃ y4567890123∀ x ∃ y4567890123∀ x ∃ y4567890123∀ x ∃ y45…
  sentence:  Dolor et alias culpa quod ab maiores tenetur aut expedita nisi v…
```

I haven't tested this on Atom or Jupyter as I don't have them installed currently. As long as they give sensible values for `displaysize(p.output)`, I hope they would also "just work".

I also changed the `print` statements in `printvalues!` to `printover` as that seemed more suitable.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -198,10 +198,14 @@ clear_ijulia() = (IJULIABEHAVIOR[] != IJuliaAppend) && isdefined(Main, :IJulia) 
 
 # update progress display
 function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, valuecolor = :blue, offset::Integer = p.offset, keep = (offset == 0), desc = p.desc)
+    if desc != p.desc
+        p.desc = desc
+        if p.barlen !== nothing
+            p.barlen += length(p.desc) - length(desc) #adjust bar length to accomodate new description
+        end
+    end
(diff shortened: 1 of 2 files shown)


ianshmean:
It'd be good to give this a try. Seems like good functionality to add.
@pengwyn, would you mind rebasing it? There's been a few changes since

pengwyn:
Done! Allowing for the new updated description too. I included another test for the dynamic resizing too.

HenriDeh:
What's the status on this PR ? I'm eager to have this functionality to avoid this 
image

timholy:
This looks great, thanks @pengwyn!

Pro tip for the interested: in the last several months I've learned that `displaysize` is not inferrable for many abstract IO types, and this leaves invalidation vulnerabilities if you, say, do some arithmetic on the return value. I'll throw a commit into #170 to make it `displaysize(io)::Tuple{Int,Int}`.

PR #157, opened 2020-05-25
[WIP] [RFC] add support for multiple simultaneous progressbars

MarcMush:
following my answer in #156 , this adds two types of progressbars that mimic `Progress`:

`ParallelProgress` is pretty much the example in the readme and works like that :
```julia
julia> using Distributed
julia> addprocs()
julia> @everywhere using ProgressMeter
julia> prog = ParallelProgress(10; desc="test ")
julia> pmap(1:10) do
           sleep(rand())
           next!(prog)
           return myid()
       end
```

`MultipleProgress` spawns multiple progress bars using `offset`
```julia
julia> using Distributed
julia> addprocs(2)
julia> @everywhere using ProgressMeter
julia> p = MultipleProgress(5,10; desc="global ", kws=[(desc="task $i ",) for i in 1:5])
       pmap(1:5) do x
           for i in 1:10
               sleep(rand())
               next!(p[x])
           end
           sleep(0.01)
           myid()
       end
global 100%|██████████████████████████████████████████████████████████| Time: 0:00:24
task 4 100%|██████████████████████████████████████████████████████████| Time: 0:00:05
task 5 100%|██████████████████████████████████████████████████████████| Time: 0:00:05
```
I use this to download multiple big files at once, where it is faster to download them simultaneously.
We can't just use multiple `ParallelProgress`es with multiple `RemoteChannel`s and different `offset`s because trying to update two at the same time will mess up the ouput

### DIFF
--- a/src/parallel_progress.jl
+++ b/src/parallel_progress.jl
@@ -86,10 +86,24 @@ end
     prog = MultipleProgress(lengths; kws, kw...)
 
 generates one progressbar for each value in `lengths` and one for a global progressbar
-`kw` arguments are applied on all progressbars
-kws[i] arguments are applied on the i-th progressbar
+ - `kw` arguments are applied on all progressbars
+ - `kws[i]` arguments are applied on the i-th progressbar
 
 # Example
+```jldoctest
+julia> using Distributed
+julia> addprocs(2)
+julia> @everywhere using ProgressMeter
+julia> p = MultipleProgress(5,10; desc="global ", kws=[(desc="task \$i ",) for i in 1:5])
+       pmap(1:5) do x
+           for i in 1:10
+               sleep(rand())
+               next!(p[x])
+           end
+           sleep(0.01)
+           myid()
+       end
+```
 """
 function MultipleProgress(lengths::AbstractVector{<:Integer}; 
                           kws = [() for _ in lengths],
@@ -105,7 +119,8 @@ function MultipleProgress(lengths::AbstractVector{<:Integer};
 
     max_offsets = 1
 
-    # we must make sure that 2 progresses aren't updated at the same time
+    # we must make sure that 2 progresses aren't updated at the same time, 
+    # that's why we use only one Channel
     @async begin
         while true
             
--- a/src/parallel_progress.jl
+++ b/src/parallel_progress.jl
@@ -1,0 +1,157 @@
+"""
+`prog = ParallelProgress(n; kw...)`
+
+works like `Progress` but can be used from other workers
+Extra arguments after `update` or `cancel` are ignored
+
+# Example:
+```jldoctest
+julia> using Distributed
+julia> addprocs()
+julia> @everywhere using ProgressMeter
+julia> prog = ParallelProgress(10; desc="test ")
+julia> pmap(1:10) do
+           sleep(rand())
+           next!(prog)
+           return myid()
+       end
+```
+"""
+struct ParallelProgress{C}
+    channel::C
+    n::Int
+end
+
+const PP_NEXT = -1
+const PP_FINISH = -2
+const PP_CANCEL = -3
+
+next!(pp::ParallelProgress) = put!(pp.channel, PP_NEXT)
+finish!(pp::ParallelProgress) = put!(pp.channel, PP_FINISH)
+cancel(pp::ParallelProgress, args...; kw...) = put!(pp.channel, PP_CANCEL)
+update!(pp::ParallelProgress, counter, color = nothing) = put!(pp.channel, counter)
+
+function ParallelProgress(n::Int; kw...)
+    channel = RemoteChannel(() -> Channel{Int}(n))
+    progress = Progress(n; kw...)
+    
+    @async while progress.counter < progress.n
+        f = take!(channel)
+        if f == PP_NEXT
+            next!(progress)
+        elseif f == PP_FINISH
+            finish!(progress)
+            break
+        elseif f == PP_CANCEL
+            cancel(progress)
+            break
+        elseif f >= 0
+            update!(progress, f)
+        end
+    end
+    return ParallelProgress(channel, n)
+end
+
+struct MultipleChannel{C}
+    channel::C
+    id
+end
+Distributed.put!(mc::MultipleChannel, x) = put!(mc.channel, (mc.id, x))
+
+
+struct MultipleProgress{C}
+    channel::C
+    amount::Int
+    lengths::Vector{Int}
+end
+
+Base.getindex(mp::MultipleProgress, n::Integer) = ParallelProgress(MultipleChannel(mp.channel, n), mp.lengths[n])
+finish!(mp::MultipleProgress) = put!.([mp.channel], [(p, PP_FINISH) for p in 1:mp.amount])
+
+
+"""
+    prog = MultipleProgress(amount, lengths; kw...)
+
+equivalent to 
+
+    MultipleProgress(lengths*ones(T,amount); kw...)
+
+"""
+function MultipleProgress(amount::Integer, lengths::T; kw...) where T <: Integer
+    MultipleProgress(lengths*ones(T,amount); kw...)
+end
+
+
+"""
+    prog = MultipleProgress(lengths; kws, kw...)
+
+generates one progressbar for each value in `lengths` and one for a global progressbar
+`kw` arguments are applied on all progressbars
+kws[i] arguments are applied on the i-th progressbar
+
+# Example
+"""
+function MultipleProgress(lengths::AbstractVector{<:Integer}; 
+                          kws = [() for _ in lengths],
+                          kw...)
+    @assert length(lengths) == length(kws) "`length(lengths)` must be equal to `length(kws)`"
+    amount = length(lengths)
+
+    total_length = sum(lengths)
+    main_progress = Progress(total_length; offset=0, kw...)
+    progresses = Union{Progress,Nothing}[nothing for _ in 1:amount]
+    taken_offsets = Set(Int[])
+    channel = RemoteChannel(() -> Channel{Tuple{Int,Int}}(max(2amount, 64)))
+
+    max_offsets = 1
+
+    # we must make sure that 2 progresses aren't updated at the same time
+    @async begin
+        while true
+            
+            (p, value) = take!(channel)
+
+            # first time calling progress p
+            if isnothing(progresses[p])
+                # find first available offset
+                offset = 1
+                while offset in taken_offsets
+                    offset += 1
+                end
+                max_offsets = max(max_offsets, offset)
+                progresses[p] = Progress(lengths[p]; offset=offset, kw..., kws[p]...)
+                push!(taken_offsets, offset)
+            end
+
+            if value == PP_NEXT
+                next!(progresses[p])
+                next!(main_progress)
+            else
+                prev_p_value = progresses[p].counter
+                
+                if value == PP_FINISH
+                    finish!(progresses[p])
+                elseif value == PP_CANCEL
+                    cancel(progresses[p])
+                elseif value >= 0
+                    update!(progresses[p], value)
+                end
+
+                update!(main_progress, 
+                        main_progress.counter - prev_p_value + progresses[p].counter)
+            end
+
+            if progresses[p].counter >= lengths[p]
+                delete!(taken_offsets, progresses[p].offset)
+            end
+
+            main_progress.counter >= total_length && break
+        end
+
+        print("\n" ^ max_offsets)
+    end
+
+    return MultipleProgress(channel, amount, collect(lengths))
+end
+
+
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -854,4 +854,6 @@ function ncalls(mapfun::Function, map_args)
     end
 end
 
+include("parallel_progress.jl")
+
 end
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -3,7 +3,9 @@ module ProgressMeter
 using Printf: @sprintf
 using Distributed
 
-export Progress, ProgressThresh, ProgressUnknown, BarGlyphs, next!, update!, cancel, finish!, @showprogress, progress_map, progress_pmap, ijulia_behavior
+export Progress, ProgressThresh, ProgressUnknown, BarGlyphs, next!, update!, cancel, 
+       finish!, @showprogress, progress_map, progress_pmap, ijulia_behavior,
+       MultipleProgress, ParallelProgress
 
 """
 `ProgressMeter` contains a suite of utilities for displaying progress


ianshmean:
This is great! It's exactly what I was just looking for, and seems stable on a few trials across remote distributed workers.

One minor thought, I noticed that the `color` field of the added `update!()` method isn't used. It might be nice to be able to change that on the fly. I couldn't see an easy way to implement that with this approach. 
I've also in the past updated the progress bar description during runs by changing `prog.desc` directly. It might be nice to add that functionality, but that could definitely wait for a future PR.

MarcMush:
I added tests and fixed some bugs due to overshooting

@ianshmean it wouldn't be difficult to change it, by passing `NamedTuple`s instead of `Int`s in the `RemoteChannel`

MarcMush:
i'm working on it here https://github.com/MarcMush/ProgressMeter.jl/tree/parallel_update

jtrakk:
This looks really nice. Does it work for tasks distributed over threads as well?

MarcMush:
yes, it should work, you can even use it on a single core
I didn't work on this much since I did this PR, but I can continue working on it if there is demand for it (or if someone wants to continue it, feel free to do it)

jtrakk:
I think it would be really useful for parallel long-running tasks. For example, it would be a nice improvement in the UX of sampling for parallel algorithms in Turing.jl.

stevengj:
It would be nicer if you could encapsulate any other progress bars, e.g. `ProgressUnknown` etcetera, in this way.  That is, if you could just pass an array of progress meters:
```
p = MultipleProgress([Progress(...), ProgressUnknown(...), ProgressThresh(...)])
```

bdeonovic:
Whats the status of this?

MarcMush:
I did some work on this PR
I implemented @IanButterworth's suggestion https://github.com/timholy/ProgressMeter.jl/pull/157#issuecomment-638330463

MarcMush:
> It would be nicer if you could encapsulate any other progress bars, e.g. `ProgressUnknown` etcetera, in this way. That is, if you could just pass an array of progress meters:
> 
> ```julia
> p = MultipleProgress([Progress(...), ProgressUnknown(...), ProgressThresh(...)])
> ```

I really like the idea, and I enabled it for `ParallelProgress`, but for `MultipleProgress`, there will be a problem for tasks that are not started immediatly, since the start time will report when the whole object was created
There is several solutions to this, I don't know which one is better:
- just don't care, the timings will be wrong in some cases
- pass functions that will be executed at first call of the progressmeter eg 
`p = MultipleProgress([()->Progress(...), ()->ProgressUnknown(...), ()->ProgressThresh(...)])`
- reset the `p.tinit` at first call

note that right now, and with one these solutions (except the first one) the first step won't be included in the timing except if initialized with `update!(p,0)` for example

stevengj:
> but for `MultipleProgress`, there will be a problem for tasks that are not started immediately, since the start time will report when the whole object was created

How about have a `Channel` object that you can `put!` progress bars to asynchronously?   That way you don't need to know in advance how many progress bars you need (e.g. if you are recursively processing a directory tree).

jtrakk:
It would be cool to have a full tree of hierarchical progress bars for hierarchical task trees, matching up with the move toward [structured concurrency](https://vorpus.org/blog/notes-on-structured-concurrency-or-go-statement-considered-harmful/). https://github.com/JuliaLang/julia/issues/33248

MarcMush:
*again*, a lot of changes
this includes #215, which is of course necessary for this

Now, there are 2 main ways to use  a MultipleProgress, with a few options
```julia
MultipleProgress([Progress(10), ProgressUnknown(), ProgressTresh(0.1)...], [mainprogressbar])

p = MultipleProgress(ProgressUnknown())
addprogress!(p[1], Progress, 10, desc="task 1")
```

I tried to detail the different possibilities in the docstring and in the tests

src/parallel_progress.jl
```diff
@@ -0,0 +1,376 @@
...
+ - `auto_reset_timer`: tinit in progresses will be reset at first call
+
+use p[i] to access the i-th progressmeter, and p[0] to access the main one
+
+# Example
+```jldoctest
+julia> using Distributed
+julia> addprocs(2)
+julia> @everywhere using ProgressMeter
+julia> p = MultipleProgress(fill(10,5); desc="global ", kws=[(desc="task \$i ",) for i in 1:5])
```
stevengj, line 137: `fill(10,5)` doesn't match the signature in the docstring?

src/parallel_progress.jl
```diff
@@ -0,0 +1,376 @@
...
+
+"""
+    MultipleProgress(progresses::AbstractVector{<:AbstractProgress}; count_finishes=false, kwmain=(), kw...)
+
+is equivalent to:
+
+    MultipleProgress(progresses, prog; count_finishes, kw...)
+
+with `prog` a `Progress` of the same length as `progresses` if `count_finishes == true`, 
+otherwise, `prog` is a `ProgressUnknown`
```
stevengj, line 189: Couldn't this be documented in the form of a default argument `mainprogresses=Progress(...)` in the main docstring, rather than as a separate docstring?

src/parallel_progress.jl
```diff
@@ -0,0 +1,376 @@
...
+end
+
+"""
+    MultipleProgress(mainprogress::AbstractProgress; auto_close=false, kw...)
+
+is equivalent to
+
+    MultipleProgress(AbstractProgress[], mainprogress; auto_close, kw...)
+"""
+function MultipleProgress(mainprogress::AbstractProgress; auto_close=false, kw...)
```
stevengj, line 207: Should `mainprogress` default to `ProgressUnknown` in this case?

src/parallel_progress.jl
MarcMush: yes, this doesn't work anymore, I'll fix the docstring

src/parallel_progress.jl
MarcMush: yes, I've added this

src/parallel_progress.jl
MarcMush: I'm not very good at documenting things, I've merged it in the main docstring like you suggested, but I don't know if it's very clear

```
- `mainprogress`: main progressbar, defaults to `Progress` or `ProgressUnknown`,
 according to `count_finishes` and whether all progresses have known length or not
```

src/parallel_progress.jl
```diff
@@ -0,0 +1,364 @@
...
+has_finished(p::ProgressUnknown) = p.done
+has_finished(p::ParallelProgress) = isfakechannel(p.channel)
+has_finished(p::MultipleProgress) = isfakechannel(p.channel)
+
+isfakechannel(_) = false
+isfakechannel(::FakeChannel) = true
+isfakechannel(mc::MultipleChannel) = isfakechannel(mc.channel)
+
+"""
+    addprogress!(mp[i], T::Type{<:AbstractProgress}, args...; kw...)
```
stevengj, line 342: Shouldn't you be able to just do `addprogress!(mp, T, args...)` and have it add the progress bar at the end?

Also, this function should probably return `p[i]`.   That way you can do `pcur = addprogress(...)` and then call `next!(pcur)`.

src/parallel_progress.jl
```diff
@@ -0,0 +1,364 @@
...
+
+will add the progressbar `T(args..., kw...)` to the MultipleProgress `mp` at index `i`
```
stevengj, line 344: To be clear, if there was already a progress bar `mp[i]`, what happens to it?  Does it get shifted to `i+1`?

stevengj: It's a bit confusing that you modify `mp` by passing one of its *elements* to this function. 

Wouldn't it be clearer to use `insert!` etcetera?  e.g.
```jl
insert!(mp, i, T, args...) # inserts at index i
push!(mp, T, args...) # adds to end
pushfirst!(mp, T, args...) # inserts at beginning
```
?   These semantics would require the return value to be `mp`.   Then you would use `mp[i]`, `mp[end]`, and `mp[begin]` to get the corresponding progress bar.

src/parallel_progress.jl
```diff
@@ -0,0 +1,364 @@
...
+end
+Distributed.put!(mc::MultipleChannel, x) = put!(mc.channel, (mc.id, x...))
+
+mutable struct MultipleProgress
+    channel
+    amount::Int
+end
+
+Base.getindex(mp::MultipleProgress, n) = ParallelProgress.(MultipleChannel.(Ref(mp.channel), n))
+Base.lastindex(mp::MultipleProgress) = mp.amount
```
stevengj, line 108: You should also have `Base.firstindex(mp::MultipleProgress) = 1` and maybe support the iteration protocol?

stevengj: Should it be a subtype of `AbstractVector{ParallelProgress}`?   That way you get iteration etcetera for free (once you implement `size`)

MarcMush: I thought about that and wasn't sure how to do it. Right now `mp[0]` is the main progressbar, which means if you want to update all progressbars without the main one, you do `mp[1:end]` and if you want to include the main one you can do `mp[0:end]`.

src/parallel_progress.jl
MarcMush: right now the new progressbar would take the place of the previous and if it isn't finished, it would stay stuck. I was thinking of a usecase where each worker has only one task and they can all do `mp[myid()]`

`push!` is probably the best, since we don't want the other progressbars changing index
and/or use `setindex!` or `replace!` to be more explicit that it replaces the previous progressbar

MarcMush: concerning `push!`, I was worried two workers would take the same index at the same time, I'm not sure how to make sure it doesn't happen

MarcMush: `mp.amount` isn't properly updated when there is a new progressbar, because each worker has the state of `mp` when the task was spawned. They can't know if other workers have created new progressbars

I see two possible way of doing this
- use randomized ids for the progressbars, with low enough probability of collision, but I don't know how the interface `mp[i]` would work
- use some sort of second channel or another way to update about `mp.amount`

src/parallel_progress.jl
```diff
@@ -0,0 +1,364 @@
...
+    id
+end
+Distributed.put!(mc::MultipleChannel, x) = put!(mc.channel, (mc.id, x...))
+
+mutable struct MultipleProgress
+    channel
+    amount::Int
+end
+
+Base.getindex(mp::MultipleProgress, n) = ParallelProgress.(MultipleChannel.(Ref(mp.channel), n))
```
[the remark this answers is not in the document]
MarcMush: `MultipleProgress`, `ParallelProgress` and `MultipleChannel` are just wrappers around a channel, they are pretty light-weight I think, but I have no idea what kind of overhead this is. 
I found it odd to store an array of redundant `ParallelProgress`es, which would be sent to all workers

I can't think of a better way of doing it, I'm open to suggestions

the user can still "cache" it themself:
```julia
p = mp[i]
for i in 1:100 next!(p) end
```

src/parallel_progress.jl
stevengj: I doubt that progressbars should be referred to using `mp[i]` at all, because of the possibility of race conditions and collisions as progressbars are added removed.

For that reason, maybe it's best to not use `push!` semantics (returning the collection), and instead stick with `p = addprogress!(mp, Type, args...)` returning the new progress bar (appended at the end)?

Why can't you do `p = addprogress!(mp, Type(args...))`, by the way?

MarcMush: > Why can't you do p = addprogress!(mp, Type(args...)), by the way?

all functions actually do a `put!(channel, (progressid, action, args, kwargs))` for the main worker to process and apply the right function
`Progress[something]` created on one worker and called with `next!` on another doesn't work
this could be done with a function or a macro I guess
or create a new progress with all the same parameters

MarcMush: > I doubt that progressbars should be referred to using mp[i] at all, because of the possibility of race conditions and collisions as progressbars are added removed.

when defining the progresses before, it makes sense to use `mp[i]`: (this works right now)

```julia
ntasks = 10
mp = MultipleProgress([Progress(100) for _ in 1:ntasks])
@sync for task in 1:ntasks
    @async for i in 1:100
        next!(mp[task])
        sleep(rand()/10)
    end
end
```

but when the progresses are defined in the task, it makes more sense like you say: (this doesn't work right now)

```julia
ntasks = 10
mp = MultipleProgress()
@sync for task in 1:ntasks
    p = addprogress!(mp, Progress, 100)
    @async for i in 1:100
        next!(p)
        sleep(rand()/10)
    end
end
finish!(mp)
```

I don't know if we should support both or if they should be 2 different structs

MarcMush: @stevengj I don't really know how to implement this...

 I agree it would be nice to not have to specify an `id` and do 
```julia
p = addprogress!(...)
```

but we need to know the id of the newly created progress for `next!`, `update!` etc.

I had 3 ideas that all have issues:
- generate a random id (for example `rand(Int)`) but it can be easily broken with `@everywhere seed!(1)`. I don't know rng enough to know if this can be fixed with some sort of `rand(rng, Int)`
- have a second `RemoteChannel` so that the main worker gives the id back to the working worker. but I don't think we can make sure the right worker `take!`s the right id when 2 progresses are created at the same time
- include `myid()` in the id, but this will fail if a single worker spawns multiple progresses at the same time with `@async` for example

MarcMush:
this is basically working but there are some decisions to make regarding syntax and functionnalities

MarcMush:
This PR is about to celebrate its third anniversary, and I would love to merge this one day. Here is an attempt to list what is currently blocking to do so:

- [ ] are we satisfied with the current `addprogress!(mp[i], Progress, args...)`? This can be confusing and the index `i` needs to be known and managed correctly. See discussion https://github.com/timholy/ProgressMeter.jl/pull/157#pullrequestreview-758944656

- [ ] `mp[0]` is currently the main progressmeter, and `mp[1:end]` all the other ones. It is not an `AbstractArray`, no `firstindex`. Should this be changed?

- [ ] is everything ok with the approach? in particular the `Channel`s 

- [ ] documentation/readme

I encourage everyone interested in having multiple progressbars to try this PR with `]add ProgressMeter#a5c5b2c` and give feedback and suggestions. Everything *should* be explained in `?MultipleProgress`.

timholy:
Sorry I've not come around to reviewing this impressive PR. My main thought is that if progress bars might be added or removed at various times, integer-indexing seems like a bad idea. I'd rather use a Dict-style way of indexing, where each progress bar gets a unique key (and trying to create a new progress bar using an existing key throws an error). Keys should live for the entire lifetime of the main progress bar, even if they've `finish`ed.

MarcMush:
I've changed to a Dict for storing the progressbars and throws an error when trying to re-add a progess with the same id. It is possible to use Ints or any other type for indexing. The main progress can be specified with the kwarg `main` and recalled with `p.main`

Moelf:
this seems actually ready? gentle bump for a nod and we can do documentation etc. and ship it?

MarcMush:
have you tested it? I'd appreciate other people trying it and giving feedback, see if it is actually what is wanted
you can use it with `]add ProgressMeter#3156d04`

MarcMush:
`@distributed pmap()` was already implemented before

this PR adds `next!()` accross workers with `ParallelProgress`, as well as multiple progressbars at once with `MultipleProgress`

Moelf:
>@distributed pmap() was already implemented before

I have multiple sub tasks and also `pmap()` is not good enough because workers are unstable

MarcMush:
great, you can use this PR with `]add ProgressMeter`**`#3156d04`** and see if it works as expected for your usecase

MarcMush:
I changed the syntax for adding new progressbars, now it simply works with `setindex!`:

```julia
mp = MultipleProgress()
mp[1] = Progress(10)
next!(p[1])
```

PR #159, opened 2020-05-31
Update the description from update!()

ianshmean:
This does two things:
- Adds the option to update `desc` via `update!()`
- Provides default values for `val`, `color` and  `counter` within `update!()` so that you can just update specific kwargs without having to provide the existing `val` `color` and  `counter`, i.e. `update!(p, desc="New desc")` or `update!(p, color=:red)`


timholy:
> Edit: I think perhaps color should be a kwarg with this new setup, but that would be a breaking change

Maybe file as a new issue? At some point we'll have to go to PM2.0.

Issue #162, opened 2020-06-15
Warning in Jupyter Notebook results in bad user experience


saschatimme:
Running a computation with ProgessMeter in a Jupyter notebook now results in this warning message for every run:
```
┌ Warning: ProgressMeter by default refresh meters with additional information in IJulia via `IJulia.clear_output`, which clears all outputs in the cell. 
│  - To prevent this behaviour, do `ProgressMeter.ijulia_behavior(:append)`. 
│  - To disable this warning message, do `ProgressMeter.ijulia_behavior(:clear)`.
```

The suggestions are actionable when `ProgressMeter` is a direct dependency of the current environment. However, if the `ProgessMeter` is used in a package (let's say `MyPkg`) and only the package is loaded in the current session then `ProgressMeter.ijulia_behavior(:clear)` as well as `import ProgressMeter` both result in an error. Only `MyPkg.ProgressMeter.ijulia_behavior(:clear)` yields the desired behaviour.

timholy:
xref #147

PR #163, opened 2020-06-19
Change default IJulia behaviour

saschatimme:
This resolves the problem described in #162.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -225,7 +225,7 @@ end
 # Package level behavior of IJulia clear output
 @enum IJuliaBehavior IJuliaWarned IJuliaClear IJuliaAppend
 
-const IJULIABEHAVIOR = Ref(IJuliaWarned)
+const IJULIABEHAVIOR = Ref(IJuliaClear)
 
 function ijulia_behavior(b)
     @assert b in [:warn, :clear, :append]


xukai92:
As it was explained in https://github.com/timholy/ProgressMeter.jl/pull/147#issuecomment-614663105, the reason the default is warning is because this behaviour can sliently clear other possible outputs in the cell. Changing the default in ProgressMeter seems dangerous to me. In my opinion, either the package that uses ProgressMeter could export ProgressMeter (so that users could do the change of default as suggested by warning), or the default should be changed only within that package.

saschatimme:
I am aware that this clears other fields. But this is about trade offs and I think the much better trade off is to just clear things by default.

zsunberg:
Ugh, this is a really annoying case... It could be catastrophically confusing to someone if they have debug print statements that they don't see.

Perhaps here are a few alternate things to try:
1. Can we detect if the user has printed things? (e.g. by checking if there are more lines in the output cell than we have printed) We could not clear in that case.
2. Would it be possible to change the warning to be one line that is printed above the progress bar so that it is not so intrusive?

timholy:
Would it be OK to print the warning only once?

I'm not an IJulia user so I'll leave it up to others to decide.

xukai92:
I guess printing it only for the first time sounds the best compremise and is also doable using similar tricks here in JuliaGPU for `allowscalar`: https://github.com/JuliaGPU/GPUArrays.jl/blob/fa6e0b3ea24b4931b3223e858133f90cf019b5b8/src/host/indexing.jl#L21

timholy:
I'd support that change.

IanButterworth:
I didn't mean to rebase this branch. But while I'm here, can someone decide if this is good to go? (I don't use IJulia)

xukai92:
> I guess printing it only for the first time sounds the best compremise and is also doable using similar tricks here in JuliaGPU for `allowscalar`: https://github.com/JuliaGPU/GPUArrays.jl/blob/fa6e0b3ea24b4931b3223e858133f90cf019b5b8/src/host/indexing.jl#L21

This change doesn't seem to be implemented.

PR #164, opened 2020-06-27
make showprogress work with reduce

johnnychen94:
The current mechanism already works with `reduce` perfectly, so I just relax the assertion to allow `:reduce` and add some test cases for it.


johnnychen94:
@ianshmean would you mind taking a look at this?

PR #166, opened 2020-09-04
Add ProgressThresh constructor

josePereiro:
Hi, this is the issue
```
julia> ProgressThresh(1.0; desc = "Desc: ")
ERROR: MethodError: no method matching ProgressThresh(::Float64, ::Float64, ::String, ::Symbol, ::Base.TTY; desc="Desc: ")
[...]
```

But, the help is a little confusing here
```
help?> ProgressThresh
search: ProgressThresh

  prog = ProgressThresh(thresh; dt=0.1, desc="Progress: ", color=:green, output=stderr) creates a progress meter for a task which will terminate once a value less than or equal
  to thresh is reached. Output will be generated at intervals at least dt seconds apart, and perhaps longer if each iteration takes longer than dt. desc is a description of the
  current task.
```

It took me a while to realize that it was because it is a parameterized type. So, this works
```
julia> ProgressThresh{Float64}(1.0; desc = "Desc: ")
ProgressThresh{Float64}(1.0, ReentrantLock(nothing...
[...]
```
There is already a constructor that allow you to use the type of `thresh`, I just add another for the kwarg version.

Thanks!!

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -135,9 +135,14 @@ mutable struct ProgressThresh{T<:Real} <: AbstractProgress
     end
 end
 
-ProgressThresh(thresh::Real, dt::Real=0.1, desc::AbstractString="Progress: ",
+ProgressThresh(thresh::Real, dt::Real, desc::AbstractString="Progress: ",
          color::Symbol=:green, output::IO=stderr;
          offset::Integer=0) =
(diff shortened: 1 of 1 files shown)


timholy:
Thanks @josePereiro! Are you OK with the implementation in #169? I'd prefer it because it doesn't further proliferate "legacy" positional argument constructors, and it centralizes construction of the parametric version.

josePereiro:
I actually do!

PR #168, opened 2020-09-08
Allow arbitrary step size for next!()

CarolusMagnus:
It is sometimes useful to increase the progress by a step size larger than one, so I added an additional argument to next!() with the default value of 1.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -332,7 +332,7 @@ end
 
 # update progress display
 """
-`next!(prog, [color], step = 1)` reports that step units of progress has been
+`next!(prog, [color], step = 1)` reports that `step` units of progress have been
 made. Depending on the time interval since the last update, this may
 or may not result in a change to the display.
 
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -330,14 +330,14 @@ or may not result in a change to the display.
 
 You may optionally change the color of the display. See also `update!`.
 """
-function next!(p::Union{Progress, ProgressUnknown}, step::Int = 1; options...)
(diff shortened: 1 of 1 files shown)


src/ProgressMeter.jl
```diff
@@ -324,23 +324,23 @@ end
 
 # update progress display
 """
-`next!(prog, [color])` reports that one unit of progress has been
+`next!(prog, [color], step = 1)` reports that step units of progress has been
```
timholy, line 327:
```suggestion
`next!(prog, [color], step = 1)` reports that `step` units of progress have been
```

timholy (commented):
I like it. My main concern is whether people will also want `next!(p, step)` without the `color` option. Should we instead use a keyword for this?

src/ProgressMeter.jl
martinholters, line 327:
```suggestion
`next!(prog, [color], step = 1)` reports that `step` units of progress have been
```

timholy: Jinx!

CarolusMagnus: I also changed the definition of the next!() without color argument. As Symbol != Int I do not see a possible conflict. If you like the keyword approach more I would also be perfectly happy with that.

timholy: Right, there isn't a possible conflict, but then someone might implement a positional version where `step` comes second and it supplies the default for `color`. So that's an extra method, one that introduces two places we have to have the default `color` value hard-coded.

CarolusMagnus:
I changed the step argument now to a keyword argument. I hope that is OK now.

timholy:
Thanks, @CarolusMagnus!

PR #169, opened 2020-09-09
Nonparametric constructor for ProgressThresh

timholy:
Replaces #166

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -134,13 +134,15 @@ mutable struct ProgressThresh{T<:Real} <: AbstractProgress
         new{T}(thresh, reentrantlocker, dt, typemax(T), 0, false, tfirst, tlast, printed, desc, color, output, 0, offset)
     end
 end
+ProgressThresh(thresh::Real; kwargs...) = ProgressThresh{typeof(thresh)}(thresh; kwargs...)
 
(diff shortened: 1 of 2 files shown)


src/ProgressMeter.jl
```diff
@@ -134,13 +134,15 @@ mutable struct ProgressThresh{T<:Real} <: AbstractProgress
         new{T}(thresh, reentrantlocker, dt, typemax(T), 0, false, tfirst, tlast, printed, desc, color, output, 0, offset)
     end
 end
+ProgressThresh(thresh::Real; kwargs...) = ProgressThresh{typeof(thresh)}(thresh; kwargs...)
 
-ProgressThresh(thresh::Real, dt::Real=0.1, desc::AbstractString="Progress: ",
+# Legacy constructor calls
+ProgressThresh(thresh::Real, dt::Real, desc::AbstractString="Progress: ",
          color::Symbol=:green, output::IO=stderr;
          offset::Integer=0) =
```
MarcMush, line 142: `offset` should either be a `Int64` or allow `offset::Integer` at line 130

timholy: Nice catch! But I'm ahead of you this time, I have a 2nd PR queued locally that fixes up some of these type issues that I'll submit once this merges. This just continues errors we already have in the code base.

PR #174, opened 2020-09-20
Fix bar length update

### LINKED ISSUE #173: Bar length is not updated correctly when `desc` is changed
`barlen` is not updated correctly when `desc` is changed.

MWE:

```
using ProgressMeter

p = Progress(100, desc="a", barlen=62)
for i in 1:100
    sleep(0.1)
    next!(p, desc="a"^rand(1:10))
end
```

mtsch:
Fixes #173


timholy:
Thanks!

Issue #171, opened 2020-09-20
ProgressMeter makes loops run slower


mtsch:
Adding a progress meter to a tight loop can make it run way slower. The problem is that the bar length is calculated on every call to `updateProgress`.

Here's an MWE:

```
using BenchmarkTools
using ProgressMeter

function prog_perf(n)
    prog = Progress(n)
    x = 0.0
    for i in 1:n
        x += rand()
        next!(prog)
    end
    return x
end

function noprog_perf(n)
    x = 0.0
    for i in 1:n
        x += rand()
    end
    return x
end

@btime prog_perf(1_000_000)
#  963.465 ms (2000457 allocations: 61.06 MiB)
@btime noprog_perf(1_000_000)
#  5.401 ms (0 allocations: 0 bytes)
```

This case is pretty extreme but I actually got in a situation where adding a progress meter increased the running time from ~5s to ~25s.

timholy:
Oh dear, this wasn't supposed to happen. Thanks for catching it and for the fix!

PR #175, opened 2020-09-21
Add a performance test

timholy:
To catch problems like #171.

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -12,3 +12,26 @@
 
 @test ProgressMeter.Progress(5, "Progress:", Int16(5)).offset == 5
 @test ProgressMeter.ProgressThresh(0.2, "Progress:", Int16(5)).offset == 5
+
+# Performance test (from #171)
+function prog_perf(n)
+    prog = Progress(n)
+    x = 0.0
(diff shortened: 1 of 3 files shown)


test/core.jl
```diff
@@ -12,3 +12,26 @@
...
+    x = 0.0
+    for i in 1:n
+        x += rand()
+    end
+    return x
+end
+
+prog_perf(1)
+noprog_perf(1)
+@test @elapsed(prog_perf(10^7)) < 40*@elapsed(noprog_perf(10^7))
```
garrison, line 37: I had to change this constant from `40` to `50` to get the tests to pass on my laptop with the current `release-1.6` branch of julia.  On julia `1.5.3` the tests pass without this change.  Should this constant be adjusted, or does this indicate something wrong to you?

timholy: Yes, something seems wrong. For me, Julia 1.5 runs in about 70% of the time of Julia master.

timholy: It looks mostly like a locking issue. If you profile, set the interval to 0.01s (the default of 0.001s interferes too much).

Issue #177, opened 2020-12-10
Reqest: allow for a pipe at end of map do block


fredcallaway:
I often have a structures like this in my code
```
results = map(jobs) do job
    job => work(job)
end |> Dict
```

If I try to throw a @showprogress on this, I get an error " ArgumentError: Final argument to @showprogress must be a for loop, comprehension, map, or pmap".  This can of course be fixed with a temporary variable, but if it's an easy fix to parse the do block here correctly, it would be more convenient.  More importantly, it would prevent confusion from aa somewhat mysterious error message.

timholy:
If that can be made to work without breaking existing functionality, I'd be happy to have it. Want to submit a PR?

timholy:
Closed by #179. (Pro tip: had you written "fixes #177" or "closes #177" instead of "addresses #177", this would have happened automatically.)

PR #179, opened 2020-12-13
Allow pipes after a map

fredcallaway:
Adresses #177


timholy (approved):
I'll leave this open in case anyone else wants to review it. If no complaints have emerged in a couple of days, let's merge this. (If I forget, please bump.)

fredcallaway:
My first contribution to a Julia package! :)

timholy:
And a very fine one indeed!

timholy:
It's already out in the wild, too: version 1.4.1 was released earlier today with this change. Nice feeling, isn't it?

PR #180, opened 2020-12-13
Add CompatHelper

timholy:


timholy:
Currently has no non-Julia deps, so let's save the CO2.

PR #182, opened 2021-01-06
Allow passing a function for lazy-showvalues

josePereiro:
Hi, 
`update!` and `next!` are methods so its arguments will be evaluated in every call. 
Sometimes I want to show values that are computed on the fly, and it is a waste of time to
 compute them if the progress won't be reprinted. 
So I just add a version of `printvalues!` that takes a zero-argument function as `showvalues`
so it will be evaluated just in case of re-printing.

An illustrative example
```julia
@info "Current: showvalue evaluates each iter"
@time let
    prog = Progress(10; dt = 1.0);
    for i = 1:10

        # Computation
        sleep(3/10) # add 3 seconds
        next!(prog; showvalues = [
                ("heavy val", (sleep(3/10); i)) # adds 3 additional seconds
            ]
        )
    end
    finish!(prog)
end
#= 
[ Info: Current: showvalue evaluates each iter
Progress: 100%|█████████████████████████████| Time: 0:00:06
  heavy val:  10
  6.289688 seconds (156.40 k allocations: 7.480 MiB)
=#

@info "Lazy: showvalue is a function evaluated only when updated"
@time let
    prog = Progress(10; dt = 1.0);
    for i = 1:10

        # Computation
        sleep(3/10) # add to 3 seconds
        next!(prog; showvalues = () -> [
                ("heavy val", (sleep(3/10); i)) # it is evaluated ~3 times so add an extra 1 second
            ]
        )
    end
    finish!(prog)
end
#= 
[ Info: Lazy: showvalue is a function evaluated only when updated
Progress: 100%|█████████████████████████████| Time: 0:00:04
  heavy val:  10
  4.418721 seconds (439.13 k allocations: 23.509 MiB)
=#
```

Thanks!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,7 +236,8 @@ generate_showvalues(iter, x) = () -> [(:iter,iter), (:x,x)]
 for iter = 1:10
     x *= 2
     sleep(0.5)
-    ProgressMeter.next!(p; showvalues = generate_showvalues(iter, x))
+# unlike `showvalues=generate_showvalues(iter, x)()`, this version only evaluate the function when necessary
+ProgressMeter.next!(p; showvalues = generate_showvalues(iter, x))
 end**
 ```
 
--- a/README.md
+++ b/README.md
@@ -225,9 +225,7 @@ for iter = 1:10
 end
 ```
 
-In the above example `showvalues` keyword is evaluated in each iteration, regardless
-if the progress will be reprinted or not. For avoiding that and prevent overhead a
-zero-argument function can be passed that returns the `showvalues` object.
+In the above example, the data passed to `showvalues` is evaluated even if the progress bar is not updated. To avoid this unnecessary computation and reduce the overhead you can alternatively pass a zero-argument function as a callback to the `showvalues` keyword.
 
 ```julia
 x,n = 1,10
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ if the progress will be reprinted or not. For avoiding that and prevent overhead
 zero-argument function can be passed that returns the `showvalues` object.
 
 ```julia
-**x,n = 1,10
+x,n = 1,10
 p = Progress(n)
 generate_showvalues(iter, x) = () -> [(:iter,iter), (:x,x)]
 for iter = 1:10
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -225,6 +225,21 @@ for iter = 1:10
 end
 ```
 
+In the above example `showvalues` keyword is evaluated in each iteration, regardless
+if the progress will be reprinted or not. For avoiding that and prevent overhead a
+zero-argument function can be passed that returns the `showvalues` object.
+
+```julia
+**x,n = 1,10
```
johnnychen94, line 233: A typo I guess?

```suggestion
x,n = 1,10
```

README.md
```diff
@@ -225,6 +225,21 @@ for iter = 1:10
 end
 ```
 
+In the above example `showvalues` keyword is evaluated in each iteration, regardless
+if the progress will be reprinted or not. For avoiding that and prevent overhead a
+zero-argument function can be passed that returns the `showvalues` object.
```
johnnychen94, line 230:
```suggestion
In the above example, the data passed to `showvalues` is evaluated even if the progress bar is not updated. To avoid this unnecessary computation and reduce the overhead you can alternatively pass a zero-argument function as a callback to the `showvalues` keyword.
```

README.md
```diff
@@ -225,6 +225,21 @@ for iter = 1:10
...
+
+```julia
+**x,n = 1,10
+p = Progress(n)
+generate_showvalues(iter, x) = () -> [(:iter,iter), (:x,x)]
+for iter = 1:10
+    x *= 2
+    sleep(0.5)
+    ProgressMeter.next!(p; showvalues = generate_showvalues(iter, x))
```
johnnychen94, line 239:
```suggestion
# unlike `showvalues=generate_showvalues(iter, x)()`, this version only evaluate the function when necessary
ProgressMeter.next!(p; showvalues = generate_showvalues(iter, x))
```

johnnychen94 (approved):
A nice observation. I made some suggestions to the example but feel free to take or drop it.

timholy:
Thanks, @josePereiro!

PR #185, opened 2021-01-21
make output optional

jd-lara:
This PR detects if the IO is a file or TTY and prevents printing the progress bars to file in cases like when running in an HPC with a SLURM schedule that the logs get full of noisy output.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -202,11 +202,12 @@ function ijulia_behavior(b)
 end
 
 # Whether or not to use IJulia.clear_output
-clear_ijulia() = (IJULIABEHAVIOR[] != IJuliaAppend) && isdefined(Main, :IJulia) && Main.IJulia.inited
+running_ijulia_kernel() = isdefined(Main, :IJulia) && Main.IJulia.inited
+clear_ijulia() = (IJULIABEHAVIOR[] != IJuliaAppend) && running_ijulia_kernel()
 
 # update progress display
 function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, valuecolor = :blue, offset::Integer = p.offset, keep = (offset == 0), desc::Union{Nothing,AbstractString} = nothing)
-    !p.io_is_tty && return
+    (!running_ijulia_kernel() & !p.io_is_tty) && return
     if desc !== nothing
         if p.barlen !== nothing
             p.barlen += length(p.desc) - length(desc) #adjust bar length to accommodate new description
@@ -265,7 +266,7 @@ function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, v
 end
 
 function updateProgress!(p::ProgressThresh; showvalues = (), truncate_lines = false, valuecolor = :blue, offset::Integer = p.offset, keep = (offset == 0), desc = p.desc)
-    !p.io_is_tty && return
+    (!running_ijulia_kernel() & !p.io_is_tty) && return
     p.offset = offset
     p.desc = desc
     t = time()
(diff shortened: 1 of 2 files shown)


jd-lara:
There is no test for IJulia, but here is working 
image

jd-lara:
bump @timholy

ianshmean:
If this is forcing ProgressMeter to not print at all if not on a TTY, that seems too heavy-handed.
Even if it's being sent to a file, the user may want that. 

I usually just change the update rate with something like this:
```
is_logging(io) = isa(io, Base.TTY) == false || (get(ENV, "CI", nothing) == "true") 

prog = Progress(n, dt = is_logging(stdout) ? 3 : 0.5, output = stdout)
...
```
Given that most CI seems to pipe to a log file via a TTY, the ENV check is needed to detect a CI setup.

To suppress all progress logging, you could do the same with something like `is_logging(io) ? typemax(Int) : 0.5` ?

Or perhaps a more fundamental API could be added along the lines of `ProgressMeter.enabled(state::Bool)` ? So one could `ProgressMeter.enabled(is_logging(stdout) == false)`

jd-lara:
@ianshmean but the resulting output in a file of ProgressMeter is not useful in a file, specially when it adds so much noise. Not only for CI purposes (which is not the use at hand), but in general if one wants to have a clean stderr file output it will still print the progress meter into the file and I don't see the value in that. 

I could implement something in the API as you mentioned but a progress meter to file seems very odd.

ianshmean:
I think it's just not possible to make a call for all situations/use models that everyone would be happy with. 

But making switching it off easy for users would be great.

jd-lara:
@ianshmean I added the field as an option and a helper function.

src/ProgressMeter.jl
```diff
@@ -47,6 +48,9 @@ function BarGlyphs(s::AbstractString)
     return BarGlyphs(glyphs...)
 end
 
+
+is_logging(io) = isa(io, Base.TTY) == false || (get(ENV, "CI", nothing) == "true")
```
ianshmean, line 52: I don't think this function should reside in this package, but it could be given as an example.

jd-lara: resolved in 3cab602

jd-lara:
bump @IanButterworth for a review

timholy (approved):
LGTM. Weird that GitHub's "hide whitespace changes" doesn't seem to work on this PR.

src/ProgressMeter.jl
```diff
@@ -165,13 +168,14 @@ mutable struct ProgressUnknown <: AbstractProgress
     color::Symbol        # default to green
     output::IO           # output stream into which the progress is written
     numprintedvalues::Int   # num values printed below progress in last iteration
+    enabled::Bool
 end
 
-function ProgressUnknown(;dt::Real=0.1, desc::AbstractString="Progress: ", color::Symbol=:green, output::IO=stderr)
+function ProgressUnknown(;dt::Real=0.1, desc::AbstractString="Progress: ", color::Symbol=:green, output::IO=stderr, enabled = true)
```
IanButterworth, line 174:
```suggestion
function ProgressUnknown(;dt::Real=0.1, desc::AbstractString="Progress: ", color::Symbol=:green, output::IO=stderr, enabled::Bool = true)
```

src/ProgressMeter.jl
```diff
@@ -80,13 +80,14 @@ mutable struct Progress <: AbstractProgress
                       barlen=nothing,
                       barglyphs::BarGlyphs=BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',),
                       offset::Integer=0,
-                      start::Integer=0
+                      start::Integer=0,
+                      enabled = true
```
IanButterworth, line 84:
```suggestion
                      enabled::Bool = true
```

IanButterworth (approved):
I think a fair few of the changes here are formatting for line width, perhaps those are the ones that don't get hidden by the whitespace flag?

timholy:
I'll merge tomorrow (probably want to accept @IanButterworth 's suggestions first) barring objections.

timholy:
Hah, I failed to notice the inconsistency between the README and the implementation (see #187).

PR #192, opened 2021-03-31
Add an option to show per-iteration speed

lukemerrick:
*I'm pretty new to Julia (both the language and the community), so please help me understand if I'm not following best practices here.*

As someone coming from the Python ecosystem to Julia, one of the things I missed most from my Python workflows was the the functionality of the [tqdm](https://tqdm.github.io/) package. Though ProgressMeter.jl gives me a lot of what I love about tqdm, I really missed being able to see the rate at which my loops are running. Here's an example of the default tqdm output:

100%|███████████████████████████████▉| 8014/8014 [01:37<00:00, 82.29files/s]

So I got to thinking, maybe I can add this functionality, or at least the part of it I miss. And so I dug into the source and did my best to implement a non-invasive default-off `showspeed` option. For simplicity, I chose a fixed-width (11 characters wide) text format to show just the time per iteration speed with at least two digits of precision and no numbers bigger than 100. Here are some examples:

**ProgressUnknown:**
Reading entry: 10 	 Time: 0:00:01 ( 0.11  s/it)
Reading entry: 19 	 Time: 0:00:01 (54.80 ms/it)

**Progress:**
Progress: 100%|███████████████████| Time: 0:00:03 ( 3.19 ms/it)

**ProgressThresh:**
Minimizing: Time: 0:00:01 (18 iterations) (95.67 ms/it)

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -72,6 +72,7 @@ mutable struct Progress <: AbstractProgress
     numprintedvalues::Int      # num values printed below progress in last iteration
     start::Int                 # which iteration number to start from
     enabled::Bool              # is the output enabled
+    showspeed::Bool            # should the output include average time per iteration
 
     function Progress(n::Integer;
                       dt::Real=0.1,
@@ -82,7 +83,8 @@ mutable struct Progress <: AbstractProgress
                       barglyphs::BarGlyphs=BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',),
                       offset::Integer=0,
                       start::Integer=0,
-                      enabled::Bool = true
+                      enabled::Bool = true,
+                      showspeed::Bool = false,
                      )
         RUNNING_IJULIA_KERNEL[] = running_ijulia_kernel()
         CLEAR_IJULIA[] = clear_ijulia()
@@ -90,7 +92,7 @@ mutable struct Progress <: AbstractProgress
         counter = start
         tinit = tsecond = tlast = time()
         printed = false
-        new(n, reentrantlocker, dt, counter, tinit, tsecond, tlast, printed, desc, barlen, barglyphs, color, output, offset, 0, start, enabled)
+        new(n, reentrantlocker, dt, counter, tinit, tsecond, tlast, printed, desc, barlen, barglyphs, color, output, offset, 0, start, enabled, showspeed)
     end
 end
 
@@ -119,13 +121,14 @@ mutable struct ProgressThresh{T<:Real} <: AbstractProgress
     triggered::Bool
     tinit::Float64
     tlast::Float64
-    printed::Bool        # true if we have issued at least one status update
-    desc::String         # prefix to the percentage, e.g.  "Computing..."
-    color::Symbol        # default to green
-    output::IO           # output stream into which the progress is written
+    printed::Bool           # true if we have issued at least one status update
+    desc::String            # prefix to the percentage, e.g.  "Computing..."
+    color::Symbol           # default to green
+    output::IO              # output stream into which the progress is written
     numprintedvalues::Int   # num values printed below progress in last iteration
     offset::Int             # position offset of progress bar (default is 0)
-    enabled::Bool             # is the output a file or not
+    enabled::Bool           # is the output enabled
+    showspeed::Bool         # should the output include average time per iteration
 
     function ProgressThresh{T}(thresh;
                                dt::Real=0.1,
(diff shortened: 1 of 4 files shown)


src/ProgressMeter.jl
```diff
@@ -163,19 +167,20 @@ mutable struct ProgressUnknown <: AbstractProgress
...
-    color::Symbol        # default to green
-    output::IO           # output stream into which the progress is written
+    printed::Bool           # true if we have issued at least one status update
+    desc::String            # prefix to the percentage, e.g.  "Computing..."
+    color::Symbol           # default to green
+    output::IO              # output stream into which the progress is written
     numprintedvalues::Int   # num values printed below progress in last iteration
-    enabled::Bool
+    enabled::Bool           # is the output enabled
+    showspeed::Bool         # should the output include average time per iteration
```
johnnychen94, line 176: It's not clear to me why these existing keywords are not presented in the docstring, but I think we should try to do so; moving the comment along with the keyword as the docstring.

johnnychen94 (approved):

src/ProgressMeter.jl
lukemerrick: Just added info to the docstrings as well as two more sections in the README. Please feel free to merge if this all looks good!


## Showing average time per iteration

You can include an average per-iteration duration in your progress meter by setting the optional keyword argument `showspeed=true` when constructing a `Progress`, `ProgressUnknown`, or `ProgressThresh`.

```julia
x,n = 1,10
p = Progress(n; showspeed=true)
for iter = 1:10
    x *= 2
    sleep(0.5)
    ProgressMeter.next!(p; showvalues = [(:iter,iter), (:x,x)])
end
```

will yield something like:

```
Progress:  XX%|███████████████████████████           |  ETA: XX:YY:ZZ (12.34  s/it)
```

instead of

```
Progress:  XX%|███████████████████████████                         |  ETA: XX:YY:ZZ
```

## Conditionally disabling a progress meter

In addition to the `showspeed` optional keyword argument, all the progress meters also support the optional `enable` keyword argument. You can use this to conditionally disable a progress bar in cases where you want less verbose output or are using another progress bar to track progress in looping over a function that itself uses a progress bar.

```julia
function my_awesome_slow_loop(n: Integer; show_progress=true)
    p = Progress(n; enable=show_progress)
    for i in 1:n
        sleep(0.1)
        next!(p)
    end
end
const SHOW_PROGRESS_BARS = parse(Bool, get(ENV, "PROGRESS_BARS", "true"))
m = 100
# let environment variable disable outer loop progress bar
p = Progress(m; enable=SHOW_PROGRESS_BARS)
for i in 1:m
    # disable inner loop progress bar since we are tracking progress in the outer loop
    my_awesome_slow_loop(i; show_progress=false)
    next!(p)
end
```

lukemerrick:
Is this now good to merge? It has tests and documentation and changes none of the default behavior. If someone with write permissions could click the big green button, I'd appreciate it!

IanButterworth:
Nice! I just gave this a go, and have two observations:
```
Time: 0:01:01 ( 0.61  s/it)
```
There seems to be spurious spaces within the brackets, one at the start and two between the number and units. Probably best to keep it to `(0.61 s/it)` and make sure the width calculation updates as the number and units change.

If `it` is an acceptable abbreviation, I think it should be `it.`. I think `iter.` would be clearer but space is limited. `(0.61 s/it.)` `(0.61 s/iter.)`

lukemerrick:
> There seems to be spurious spaces within the brackets, one at the start and two between the number and units. Probably best to keep it to `(0.61 s/it)` and make sure the width calculation updates as the number and units change.

I don't like my progress bars changing length as units change (one of my pet peeves with `tqdm`) since it makes it jumpy and harder to read when your iteration speed is on the borderline between orders of magnitude. Thus I opted for a fixed-width representation at the cost of spurious whitespace (and also used a flat average instead of `tqdm`'s exponentially weighted moving average to avoid another major source of jumpiness). I felt that allowing one-character units and  numbers which would have extra leading zeros otherwise to be padded with a space didn't make things too ugly, but I can respect differing opinions. Are you hard-set on no spurious spaces?

> If `it` is an acceptable abbreviation, I think it should be `it.`. I think `iter.` would be clearer but space is limited. `(0.61 s/it.)` `(0.61 s/iter.)`

`tqdm` has close to 20k stars on GitHub and uses `it`, so I opted to copy that choice as it felt standard enough to me. Personally as a user, I never had trouble understanding `it` meant `iteration`, and it's been nice to leave as much space for the bar as possible so it can be comfortably used in 80-char terminals and such.

IanButterworth:
@lukemerrick I took the liberty to rebase this, given I caused some merge conflicts on master.

I think your points are fair, especially given this is opt-in, which gives a chance for people to give feedback in the next release.

I'll merge once tests pass

lukemerrick:
@IanButterworth 

I appreciate the rebase and merge support. Excited this will be making it in the next release :sunglasses: 

> ... a chance for people to give feedback in the next release.

Please don't hesitate to ping me if there are requests relevant to my contribution. I'm happy to contribute tweaks if there is popular demand!

PR #194, opened 2021-04-06
Improve distributed example with a long-running function instead of sleep

mirkobunse:
This PR improves the [tips for parallel programming](https://github.com/timholy/ProgressMeter.jl#tips-for-parallel-programming) section in the README.

The earlier example in this section was relying on the presence of a `sleep` call, which couldn't be replaced with an actual function out of the box. The new version is using an actual dummy function, which users can replace more easily with something useful.

I was asked to contribute this improvement [in a related issue](https://github.com/timholy/ProgressMeter.jl/issues/136#issuecomment-785020733). However, I am also thinking about implementing a distributed decorator for progress bars, so that all bar types can be used in parallel implementations out-of-the-box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,26 +267,34 @@ can be accomplished with a `RemoteChannel`:
 using ProgressMeter
 using Distributed
 
-p = Progress(10)
-channel = RemoteChannel(()->Channel{Bool}(10), 1)
+n_steps = 20
+p = Progress(n_steps)
+channel = RemoteChannel(()->Channel{Bool}(), 1)
 
-@sync begin
-    # this task prints the progress bar
+# introduce a long-running dummy task to all workers
+@everywhere long_task() = sum([ 1/x for x in 1:100_000_000 ])
+@time long_task() # a single execution is about 0.3 seconds
+
+@sync begin # start two tasks which will be synced in the very end
+    # the first task updates the progress bar
     @async while take!(channel)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -267,27 +267,35 @@ can be accomplished with a `RemoteChannel`:
 using ProgressMeter
 using Distributed
 
-p = Progress(10)
-channel = RemoteChannel(()->Channel{Bool}(10), 1)
+n_steps = 20
+p = Progress(n_steps)
+channel = RemoteChannel(()->Channel{Bool}(), 1)
```
mirkobunse, line 272: Note that the `Channel` has no buffer, so that `put!(channel, ...)` will block until the progress bar is updated; this way, the `yield()` is already included and does not need to be called separately.

IanButterworth (approved):
I think this approach enforces tighter synchronization of the processes, but I think it's not much of an issue given longer running jobs are typically distributed.

Plus, the average speed of `next!` has improved on master

1.5.0
```
julia> @btime next!($p)
Progress:   6%|██▌                                      |  ETA: 0:03:29  
198.223 ns (0 allocations: 0 bytes)
```
Master
```
julia> @btime next!($p)
Progress:   7%|██████                                                                                |  ETA: 0:00:47  
14.055 ns (0 allocations: 0 bytes)
```

PR #195, opened 2021-04-16
Base ETA off second iteration onwards for higher accuracy

IanButterworth:
Given the first iteration can be slower in julia due to compilation, it appears to be better to to base the ETA timing off the second iteration onwards.

As an example, take this which has a fake 3 second delay after the progress bar init to simulate some compilation of the loop
```julia
n = 100
p = Progress(n)
sleep(3)
for i = 1:n # 5-second loop
    sleep(0.05)
    ProgressMeter.next!(p)
end
```
On master, the first ETA to print starts at about 30 seconds, then gradually reduces linearly throughout, decrementing far faster than seconds in real time and never really giving a fair ETA.

In this PR, the first ETA is 4 seconds, and decrements immediately in-step with real-time seconds towards the final completion of 5 seconds. It never seems inaccurate

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -59,7 +59,8 @@ mutable struct Progress <: AbstractProgress
     reentrantlocker::Threads.ReentrantLock
     dt::Float64
     counter::Int
-    tfirst::Float64
+    tinit::Float64
+    tsecond::Float64           # ignore the first loop given usually uncharacteristically slow
     tlast::Float64
     printed::Bool              # true if we have issued at least one status update
     desc::String               # prefix to the percentage, e.g.  "Computing..."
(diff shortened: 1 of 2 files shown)


src/ProgressMeter.jl
```diff
@@ -206,6 +206,9 @@ clear_ijulia() = (IJULIABEHAVIOR[] != IJuliaAppend) && running_ijulia_kernel()
 # update progress display
 function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, valuecolor = :blue, offset::Integer = p.offset, keep = (offset == 0), desc::Union{Nothing,AbstractString} = nothing)
     (!running_ijulia_kernel() & !p.enabled) && return
+    if p.counter == 2
+        p.tsecond = time()
+    end
```
johnnychen94, line 211: A comment here would be helpful for future readers.

johnnychen94 (approved):

IanButterworth:
The threaded windows tests have been hanging. They were previously set to 3, but apparently the linux and windows runners have 2 cores, so perhaps the windows runners don't support hyperthreading? 

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

IanButterworth:
There's a few performance improvement PRs open now, so I'll merge to avoid things becoming stale, but hold off on a release for the moment

PR #196, opened 2021-04-17
Only poll IJulia state during init

IanButterworth:
`running_ijulia_kernel()` and `clear_ijulia()` are not inexpensive functions, and they're being polled every update.

This changes them to only be polled during creation of the progress meter, then static bools are used in their place for the rest of the run. 

I don't believe this changes any behavior in IJulia?


IanButterworth:
I'm going to merge as I have another performance optimization PR to do on top of this. I'll hold off a release until that next PR is reviewed and these are okayed

PR #197, opened 2021-04-18
Learn to be lazy with time checks. Don't lock if not threading

IanButterworth:
Building on #196 for reducing unnecessary expense for tight loops.

1. Implements lazy-smart time checking. Previously `time()` was polled for every `next!`, which is about ~30ns, so can be a bit of a slow down for tight loops.
This takes the approach of iteratively predicting the number of iterations for the first print period to be reached, then using that to defer time checking for the next loop for that number of iterations. Then each time the print occurs, the timing is corrected for any overshoot or undershoot. In testing with some debug prints it seems to stabilize quite quickly to check time 1-2 times per `dt`

2. Given locks can be expensive (~140ns [on 1.7 currently](https://github.com/JuliaLang/julia/issues/38947#issuecomment-821918316)), this changes to only locking if multiple threads are detected by the next/update functions.

ProgressMeter v1.5.0
```julia
julia> function perf(n; doprog=true)
           prog = Progress(n)
           x = 0.0
           for i in 1:n
               x += rand()
               doprog && next!(prog)
           end
           return x
       end

julia> @time perf(10^7,  doprog=true)
Progress: 100%|█████████████████████████████████████████| Time: 0:00:02
  2.265462 seconds (1.04 k allocations: 62.078 KiB)
5.002344696500369e6

julia> @time perf(10^7,  doprog=false)
  0.037642 seconds (7 allocations: 400 bytes)
4.999619221211615e6
```

This PR
```julia
julia> @time perf(10^7, 0, doprog=true)
Progress: 100%|█████████████████████████████████████████| Time: 0:00:00
prog.check_iterations = 6459310
  0.150941 seconds (113 allocations: 10.078 KiB)
4.99952291278307e6

julia> @time perf(10^7, 0, doprog=false)
prog.check_iterations = 1
  0.041455 seconds (22 allocations: 976 bytes)
5.000273824098582e6
```


test/core.jl
```diff
@@ -35,4 +35,4 @@ prog_perf(10^7)
 noprog_perf(10^7)
 @time prog_perf(10^7)
 @time noprog_perf(10^7)
-@test @elapsed(prog_perf(10^7)) < 40*@elapsed(noprog_perf(10^7))
+@test @elapsed(prog_perf(10^7)) < 8*@elapsed(noprog_perf(10^7))
```
IanButterworth, line 38: Based on local tests this could go down to 5, but CI seems closer to 6/7. 8 to be conservative

src/ProgressMeter.jl
```diff
@@ -330,19 +359,44 @@ function updateProgress!(p::ProgressUnknown; showvalues = (), truncate_lines = f
...
+            p.tlast = t + 2*(time()-t)
+            p.printed = true
+            p.prev_update_count = p.counter
+            return
+        end
+    end
+end
+
+function is_threading(p::AbstractProgress)
+    length(p.threads_used) > 1 && return true
```
johnnychen94, line 386: I'm not sure of it. Will `Threads.nthreads() == 1 && return false` a plausible shortcut?

src/ProgressMeter.jl
```diff
@@ -330,19 +359,44 @@ function updateProgress!(p::ProgressUnknown; showvalues = (), truncate_lines = f
...
-        printover(p.output, msg, p.color)
-        printvalues!(p, showvalues; color = valuecolor, truncate = truncate_lines)
-        flush(p.output)
-        # Compensate for any overhead of printing. This can be
-        # especially important if you're running over a slow network
-        # connection.
-        p.tlast = t + 2*(time()-t)
-        p.printed = true
-        return
+    if p.counter - p.prev_update_count >= p.check_iterations
```
johnnychen94, line 362: For future readability, perhaps move this check to a separate function and properly document/comment on it.

johnnychen94 (approved):
I don't know all the internal details but this change looks good to me. A nice observation!

src/ProgressMeter.jl
IanButterworth: Good idea!

IanButterworth:
I've tested this in the ways I can think might challenge it, including
- Loops that start fast and slow down
- The distributed example that has heterogenous yields

.. and I haven't seen any bad behavior.

Given a few updates on master with new features and no breaking changes to the public API, in a couple of days I'll release 1.6.0 unless anyone disagrees

cc. @timholy @johnnychen94

IanButterworth:
Ok! Yes, let's add a flag to restore the old behavior. Would you happen to have a MWE that I can test and add to the test suite?

IanButterworth:
@vancleve is this a valid MWE? It demonstrates the issue you I think you have

```julia
using ProgressMeter
n = 100
p = Progress(n)
for i in 1:n
    for j in 1:10
        next!(p, step=0, showvalues = [(:j, j)])
        sleep(1)
    end
    next!(p)
end
```

PR #204, opened 2021-05-19
Fix overprinting in ProgressUnknown

palday:
This just adds in the same logic that's present in `ProgressThresh` and `Progress`.


IanButterworth:
@palday I don't use IJulia so wouldn't know how to test this, and tests don't cover this. 
If you could add some examples of before and after, that would be helpful

PR #216, opened 2021-08-13
Remove `enabled` dependence on IJulia; fix #214

### LINKED ISSUE #214: `enabled = false` not working in IJulia
I am setting `enabled = false` but I still get a progress meter. Shouldn't it be `|` instead of `&` here?
https://github.com/timholy/ProgressMeter.jl/blob/e1c7c7127848ee270b6770e03971d326afad2b85/src/ProgressMeter.jl#L261
https://github.com/timholy/ProgressMeter.jl/blob/e1c7c7127848ee270b6770e03971d326afad2b85/src/ProgressMeter.jl#L339

halleysfifthinc:
It doesn't make sense to me that `enabled` stops working as expected in IJulia (ie Progress will always updates/prints in IJulia, regardless of `enabled` value).

Fixes #214

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -91,7 +91,6 @@ mutable struct Progress <: AbstractProgress
                       enabled::Bool = true,
                       showspeed::Bool = false,
                      )
-        RUNNING_IJULIA_KERNEL[] = running_ijulia_kernel()
         CLEAR_IJULIA[] = clear_ijulia()
         reentrantlocker = Threads.ReentrantLock()
         counter = start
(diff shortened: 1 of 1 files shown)


halleysfifthinc:
Before:
image

After:
image

martinholters (approved):
It's a bit bizarre that we went through some hoops to disable functionality in IJulia for no apparent reason. If anyone actually relied on this to disable progress metering everywhere except in IJulia, that would much better be handled by conditionally setting `enabled` by the user.

goerz:
Is there anything preventing the PR from being merged? I'm affected by #214, and I'd love to see a new release that fixes this

PR #217, opened 2021-08-16
add support for offset in ProgressUnknown

MarcMush:
#115 and #113 happened nearly at the same time, so the `offset` keyword wasn't implemented for `ProgressUnknown`

this is a ctrl+c ctrl+v of what happens for the other so hopefully it works correctly

`cancel(p::ProgressUnknown)` also didn't work because of that (I added testcases for cancel on all 3 types)


timholy (approved):
I fixed a merge conflict; if tests pass, merge when ready @MarcMush

MarcMush:
remerged because of non-passing tests, ready to merge if it passes now

timholy:
Thanks!

PR #222, opened 2021-11-09
Implemented SimpleProgress suitable for output to file

jagot:
When running on e.g. a cluster, the output is usually redirected to a
file, where the escape characters turn into a jumbled mess (see
#151). This PR implements a simple swap-in for `Progress` that instead
prints a new line after each progress update:

``` julia
N = 100
p = SimpleProgress(N)
for i = 1:N
    sleep(0.01)
    ProgressMeter.next!(p)
end
```

which gives output like this:

```
Progress:   8%  ETA: 0:00:01
Progress:  17%  ETA: 0:00:01
Progress:  26%  ETA: 0:00:01
Progress:  35%  ETA: 0:00:01
Progress:  44%  ETA: 0:00:01
Progress:  53%  ETA: 0:00:01
Progress:  62%  ETA: 0:00:00
Progress:  71%  ETA: 0:00:00
Progress:  80%  ETA: 0:00:00
Progress:  89%  ETA: 0:00:00
Progress:  98%  ETA: 0:00:00
Progress: 100% Time: 0:00:01
```

`showvalues` & `showspeed` is also supported:

``` julia
N = 100
p = SimpleProgress(N, showspeed=true)
for i = 1:N
    sleep(0.01)
    ProgressMeter.next!(p, showvalues=[(:i,i),(:x,sin(i))])
end
```

```
Progress:   8%  ETA: 0:00:01 (13.12 ms/it)
  i:  8
  x:  0.9893582466233818
Progress:  17%  ETA: 0:00:01 (12.19 ms/it)
  i:  17
  x:  -0.9613974918795568
Progress:  26%  ETA: 0:00:01 (11.97 ms/it)
  i:  26
  x:  0.7625584504796027
Progress:  35%  ETA: 0:00:01 (11.84 ms/it)
  i:  35
  x:  -0.428182669496151
Progress:  44%  ETA: 0:00:01 (11.82 ms/it)
  i:  44
  x:  0.017701925105413577
Progress:  53%  ETA: 0:00:01 (11.74 ms/it)
  i:  53
  x:  0.39592515018183416
Progress:  62%  ETA: 0:00:00 (11.71 ms/it)
  i:  62
  x:  -0.7391806966492228
Progress:  71%  ETA: 0:00:00 (11.68 ms/it)
  i:  71
  x:  0.9510546532543747
Progress:  80%  ETA: 0:00:00 (11.67 ms/it)
  i:  80
  x:  -0.9938886539233752
[...]

PR #223, opened 2021-11-09
Progressmeter for output to file

MarcMush:
this is could be an alternative to #222 
this is based on what is already done for the `keep` keyword

set `keepall=true` in `next!` or `update!` for it to take effect (only for `Progress` right now, but it can easily be adapted to `ProgressTresh` and `ProgressUnknown`)

```julia-repl
julia> io = IOBuffer()
       p = Progress(5, output=io)
       for i in 1:5
       sleep(0.2)
       next!(p, keepall=true, showvalues=[(:i,i),(:x,sin(i))])
       end
       print(String(take!(io)))

Progress:  40%|█████████████████                        |  ETA: 0:00:01
  i:  2
  x:  0.9092974268256817
Progress:  60%|█████████████████████████                |  ETA: 0:00:00
  i:  3
  x:  0.1411200080598672
Progress:  80%|█████████████████████████████████        |  ETA: 0:00:00
  i:  4
  x:  -0.7568024953079282
Progress: 100%|█████████████████████████████████████████| Time: 0:00:01
  i:  5
  x:  -0.9589242746631385
```

if there are still escape characters in the output for the color, you can wrap the io in `IOContext(io, :color => false)`

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -278,11 +278,13 @@ function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, v
                 sec_per_iter = elapsed_time / (p.counter - p.start)
                 msg = @sprintf "%s (%s)" msg speedstring(sec_per_iter)
             end
-            !CLEAR_IJULIA[] && print(p.output, "\n" ^ (p.offset + p.numprintedvalues))
-            move_cursor_up_while_clearing_lines(p.output, p.numprintedvalues)
-            printover(p.output, msg, p.color)
-            printvalues!(p, showvalues; color = valuecolor, truncate = truncate_lines)
-            if keep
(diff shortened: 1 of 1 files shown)


MarcMush:
another option could be to have an global option similar to `IJULIABEHAVIOR[]` and `CLEAR_IJULIA[]`

MarcMush:
this is a draft of an alternative to #222, it's probably not hard to finish (some copy-paste and testing). I might work on it when I have the time but if someone wants this feature faster, they can work on it

timholy:
Sorry no one has given this a review. This looks reasonable and I'd support merging it if someone added tests and docs.

MarcMush:
This is still in progress, I will continue working on it

PR #240, opened 2022-05-30
Update README.md

yakir12:
Added a 
```julia
finish!(p)
```
to the threaded examples. In my experience at least, some threaded loops do not print the total time, only the last iteration. 
While I'm sure this is not trye in all threaded cases, it seems more prudent to add this to the documentation.


timholy:
Thanks!

PR #245, opened 2022-08-05
Add output buffering to avoid flickering

ikirill:
I was getting quite a bit of flickering, especially when displaying values under the progress bar, so this patch tries to address that by putting all the output into an IOBuffer before writing it all at once.

In particular, the cursor was constantly moving back and forth (visibly!) between the beginning and the end of the line.


timholy:
Let's re-run the tests here.

PR #247, opened 2022-08-11
Structure docs more

rikhuijzer:
This PR makes the docstings more consistent with Julia conventions. Without this change, it isn't very clear what arguments are allowed from the docstring alone.


timholy:
Thanks, this is a very old package and the docstrings never got updated!

PR #261, opened 2023-03-07
print progressbar also for n=1 and n=2

### LINKED ISSUE #258: no progress printed for lengths 1 or 2
When the length of the iterator is 1 or 2 the progress bar is not printed at all:
```julia
julia>  @showprogress for i in 1:1; sleep(1) ; end

julia>  @showprogress for i in 1:2; sleep(1) ; end

julia>  @showprogress for i in 1:3; sleep(1) ; end
Progress: 100%|██████████████████████████| Time: 0:00:03
```
For consistency, I think that a progress bar should be printed in any case at the end of the loop no matter what the length of the iterator is and also no matter how large/small the `dt` option in the progress bar is.

CarloLucibello:
fix #258

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -265,8 +265,9 @@ function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, v
         p.desc = desc
     end
     p.offset = offset
+    
     if p.counter >= p.n
-        if p.counter == p.n && p.printed
+        if p.counter == p.n #&& p.printed
             t = time()
             barlen = p.barlen isa Nothing ? tty_width(p.desc, p.output, p.showspeed) : p.barlen
             percentage_complete = 100.0 * p.counter / p.n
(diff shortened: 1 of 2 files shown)


src/ProgressMeter.jl
```diff
@@ -265,8 +265,9 @@ function updateProgress!(p::Progress; showvalues = (), truncate_lines = false, v
         p.desc = desc
     end
     p.offset = offset
+    
     if p.counter >= p.n
-        if p.counter == p.n && p.printed
+        if p.counter == p.n #&& p.printed
```
CarloLucibello, line 270: @timholy I'm not sure about the effect of this. What's the purpose of `p.printed`?

CarloLucibello:
@timholy could you take a look? I run the tests locally and they seem fine by visual inspection

timholy:
The default behavior of not printing needs to be retained; users should opt-in if they want it to print. To make this customizable you'd probably have to add another field to `Progress`.

src/ProgressMeter.jl
timholy: It's part of ensuring you don't start printing only at finish: you have to print at least one update for this block to print anything at all.

timholy:
Oh, I'd forgotten https://github.com/timholy/ProgressMeter.jl/issues/258#issuecomment-1437028566, sorry.

PR #262, opened 2023-03-22
Make @showprogress support keyword arguments

### LINKED ISSUE #229: Enabled keyword for macro @showprogress
Can the macro `@showprogress` support an `enabled` boolean arg?
Just like the other functions do. I think it would be convenient.

dhanak:
This allows constructs such as:

```jl
@showprogress "X: " color=:red for x in 1:10
    sleep(1)
end
```

and

```jl
@showprogress showspeed=true for x in 1:10
    sleep(1)
end
```

Fixes #229.


timholy (approved):
Somehow the tests didn't run, I'll close and reopen.

PR #225, opened 2021-11-10
Fix unsupported keyword argument, e.g., showspeed

islent:
Fix the error:
```julia
ERROR: MethodError: no method matching Progress(::Int64, ::String; showspeed=true)
Closest candidates are:
  Progress(::Integer, ::AbstractString) at /home/islent/.julia/packages/ProgressMeter/Vf8un/src/ProgressMeter.jl:109 got unsupported keyword argument "showspeed"
  Progress(::Integer, ::AbstractString, ::Integer) at /home/islent/.julia/packages/ProgressMeter/Vf8un/src/ProgressMeter.jl:109 got unsupported keyword argument "showspeed"
  Progress(::Integer; dt, desc, color, output, barlen, barglyphs, offset, start, enabled, showspeed) at /home/islent/.julia/packages/ProgressMeter/Vf8un/src/ProgressMeter.jl:82
```

PR #275, opened 2023-09-01
Deprecate non-kwargs methods

### LINKED ISSUE #33: Allow to update the description
It would be nice, if ProgressMeter can support updating the description line during the task execution, e.g. to monitor the current minimal value during function optimization.

### LINKED ISSUE #160: Make color a kwarg in update!() (breaking change)
A note for inclusion in the next breaking release version.

https://github.com/timholy/ProgressMeter.jl/pull/159 adds the option to update `desc` as a kwarg in `update!()` and highlighted that `color` should probably become a kwarg.

i.e. if color was a kwarg, all of these would be valid
```julia
update!(prog, color=:blue)
update!(prog, desc="Step 2...")
update!(prog, offset=2)
```

Currently you have to provide counter to update the color. The other updates would work as above.

```julia
update!(prog, counter=prog.counter, color=:blue)
update!(prog, desc="Step 2...")
update!(prog, offset=2)
```

### LINKED ISSUE #165: MethodError using color kwarg
Using `ProgressMeter.jl` installed from `General` with `]add ProgressMeter`, I'm trying to follow the guide in the README, and attempting to construct a progress bar that is blue, and the code is written:
```julia
p = Progress(100, 1, color = :blue)
```
but it yields the error
```
ERROR: MethodError: no method matching Progress(::Int64, ::Int64, ::String, ::Int64, ::Symbol, ::Base.TTY; color=:blue)

Closest candidates are:
  Progress(::Integer, ::Real, ::AbstractString, ::Integer, ::Symbol, ::IO; offset) at /home/pdimens/.julia/packages/ProgressMeter/OUQkp/src/ProgressMeter.jl:95 got unsupported keyword argument "color"
  Progress(::Integer, ::Real, ::AbstractString, ::Integer, ::Symbol) at /home/pdimens/.julia/packages/ProgressMeter/OUQkp/src/ProgressMeter.jl:95 got unsupported keyword argument "color"
  Progress(::Integer, ::Real, ::AbstractString, ::Integer) at /home/pdimens/.julia/packages/ProgressMeter/OUQkp/src/ProgressMeter.jl:95 got unsupported keyword argument "color"
```
However, the docstring seems to also indicate the keyword can be used
```
help?> Progress
search: Progress progress_map ProgressMeter progress_pmap ProgressThresh ProgressUnknown @showprogress

  prog = Progress(n; dt=0.1, desc="Progress: ", color=:green, output=stderr, barlen=tty_width(desc),
  start=0) creates a progress meter for a task with n iterations or stages starting from start. Output
  will be generated at intervals at least dt seconds apart, and perhaps longer if each iteration takes
  longer than dt. desc is a description of the current task.
```
Is this a conflict of information between for another release, or perhaps I'm doing something wrong?

### LINKED ISSUE #220: Enabled keyword not in every method
For example, `Progress` defined at line 106 (i.e., the method in which you specify the minimum update interval, `dt`) fails with the `enabled` keyword.  My current workaround is
```
p = Progress(10_000_000, 1) # minimum update interval of 1 second
p.enabled = false
```

### LINKED ISSUE #254: Docs suggestion: @showprogress Inf
I often want to add the option to conditionally disable a progressbar I've defined with `@showprogress`. Previously I would always switch to progress_map and use the enabled key word, but it just occurred to me that you can accomplish this much more easily by setting the dt parameter to Inf, e.g. `@showprogress Inf map(func, args)`. Might be worth adding this as an example in the docs for other users... or maybe this is obvious and I'm just really slow :P

### LINKED ISSUE #259: Make function to change value of `n` in `Progress`
Sometimes, we want a function that gets an object of type `Progress` to be able to change the `n` parameter.  The rationale is that the caller knows better which message to show, or how long the progress bar has to be, and that the function called, which is doing all the work, knows better what `n` should be.  For example:

```
function caller()
    bar = Progress(1, desc="Working")
    f(bar)
end

function f(bar)
    bar.n = 20
    # do something, update bar, etc.
end
```

I checked that the above code works, since the constructor of `Progress` is not computing anything from `n`.  However, it would be good to have a function to change `n`, since this may change in the future.  So we could do:

```
function f(bar)
    set_max!(bar, 20)
    # do something
end
```

MarcMush:
deprecates non-kwargs methods to build progresses and `next!`/`update!`

for example:
```julia
Progress(10; desc="My progress")
ProgressUnknown(desc = "Computing...", color=:red)
next!(p; desc="New description")
update!(p; color=:blue)
```

also adds `max_steps` and `thresh` as kwargs for `next!` and `update!`. I'm not sure about the name `max_steps`, but I don't like `max` or `n`, feel free to suggest something else.

closes #33, closes #160, closes #165, closes #220, closes #225, closes #254, closes #259

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -798,7 +798,7 @@ function showprogressdistributed(args...)
 
     setup = quote
         n = length($(esc(r)))
-        p = Progress(n, $(showprogress_process_args(progressargs)...)) #? kwargs?
+        p = Progress(n, $(showprogress_process_args(progressargs)...))
         ch = RemoteChannel(() -> Channel{Bool}(n))
     end
 
@@ -842,16 +842,15 @@ end
 
 """
 ```
-@showprogress dt "Computing..." for i = 1:50
+@showprogress [desc="Computing..."] for i = 1:50
     # computation goes here
 end
 
-@showprogress dt "Computing..." pmap(x->x^2, 1:50)
+@showprogress [desc="Computing..."] pmap(x->x^2, 1:50)
 ```
-displays progress in performing a computation. `dt` is the minimum
-interval in seconds between updates to the user. You may optionally 
+displays progress in performing a computation.  You may optionally 
 supply a custom message to be printed that specifies the computation 
-being performed.
+being performed or other options.
 
 `@showprogress` works for loops, comprehensions, map, reduce, and pmap.
 """
@@ -946,7 +945,7 @@ function showprogress(args...)
 
         setup = quote
             iterable = $(esc(obj))
-            $(esc(metersym)) = Progress(length(iterable), $(showprogress_process_args(progressargs)...)) #? kwargs?
+            $(esc(metersym)) = Progress(length(iterable), $(showprogress_process_args(progressargs)...))
         end
 
         if expr.head === :for
@@ -996,7 +995,7 @@ function showprogress(args...)
 
         # create appropriate Progress expression
         lenex = :(ncalls($(esc(mapfun)), ($([esc(a) for a in mapargs]...),)))
-        progex = :(Progress($lenex, $([esc(a) for a in progressargs]...)))
+        progex = :(Progress($lenex, $(showprogress_process_args(progressargs)...)))
 
         # insert progress and mapfun kwargs
         push!(call.args, Expr(:kw, :progress, progex))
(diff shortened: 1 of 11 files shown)


src/ProgressMeter.jl
```diff
@@ -964,7 +946,7 @@ function showprogress(args...)
 
         setup = quote
             iterable = $(esc(obj))
-            $(esc(metersym)) = Progress(length(iterable), $(showprogress_process_args(progressargs)...))
+            $(esc(metersym)) = Progress(length(iterable), $(showprogress_process_args(progressargs)...)) #? kwargs?
```
MarcMush, line 949: not sure if these need to be changed (same L801)

MarcMush:
I'm really not sure about the `desc` kwarg

it's probably the most used option, so it can be annoying to add `desc=` everywhere:

```julia
Progress(10; desc="Computing...")
ProgressThresh(0.1; desc="Optimizing...")
ProgressUnknown(desc="Searching...")
@showprogress desc="Iterating..." for i in 1:1000
```

at the same time I like the symetry with `update!`/`next!`:

```julia
p = Progress(10; desc="Computing...", color=:blue)
update!(p; desc="Finishing...", color=:red)
```

(well, I guess the symetry is broken again with `thresh` and `max_steps`)

timholy:
I agree, it is commonly used. That said, I think there's a consistency argument:
- One thing you *need* to specify for `Progress` is the number of steps; it really doesn't make any sense to have a default value. Likewise with `ProgressThresh` you need to specify the threshold. But for everything else, I think there's a reasonable default.
- With `update!` the intention is to advance-to-next, so if you're modifying anything else then it should be a kwarg.

So even if it's annoying I suspect we should make `desc` required as a keyword argument. I am a bit ambivalent the name `desc`, the occasional user (e.g, me 🙂) is going to have a hard time remembering the specific abbreviation: "is it `description=` or `desc=` or something else?"

timholy (approved):
Have you checked that a version of this PR lacking the changes to `test/` also passes tests? If so, I approve merging. Many thanks!

MarcMush:
yes, the old tests pass with a bunch of deprecation warnings (I tested locally on 0.7 and 1.10)

timholy:
Great work!

MarcMush:
thanks :)

PR #278, opened 2023-09-05
ProgressMeter 1.9.0

IanButterworth:
I think https://github.com/timholy/ProgressMeter.jl/pull/217 requires a minor bump.

Happy to hold off if other things are imminent, just wanted to get fix #274 out


MarcMush:
should I merge #275 and #276 first? or should it be 1.10?

timholy:
Yeah, let's get those merged and then tagged. That said, we can merge this now if there isn't going to be a 1.8.1 (I'll hold off on hitting merge here just in case @MarcMush has other plans).

Issue #281, opened 2023-10-24
Errors in CI with threads


MarcMush:
During the CI of #157, this error happened once only (didn't happen when re-running the job)
https://github.com/timholy/ProgressMeter.jl/actions/runs/6605620538/job/17940975807?pr=157#step:7:15122

this code errored: 
https://github.com/timholy/ProgressMeter.jl/blob/5b3bd1d3ed1828a43ddd477ef085f586530c5614/test/test_threads.jl#L43-L61

basically `sum([300.0; -rand()...]) == NaN` ??

the seed is set in `test.jl` so it should be repetable

is it because of threads?  It doesn't seem to be a problem with `ProgressMeter.jl` really

Or is it not a big deal since it happened only once in a CI?

MarcMush:
I can reproduce it easily on my computer:

```julia-repl
PS C:\Users\Marc> julia -t 12 --startup-file no
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.10.0-rc1 (2023-11-03)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> vals = [1.0]
       @Threads.threads for i in 1:100000
           push!(vals, rand())
           if sum(vals) |> isnan
               error("isnan at i=$i in thread ", Threads.threadid())
           end
       end

ERROR: TaskFailedException

    nested task error: isnan at i=25079 in thread 5
    Stacktrace:
     [1] error(::String, ::Int64)
       @ Base .\error.jl:44
     [2] macro expansion
       @ .\REPL[2]:5 [inlined]
     [3] (::var"#37#threadsfor_fun#8"{var"#37#threadsfor_fun#7#9"{UnitRange{Int64}}})(tid::Int64; onethread::Bool)
       @ Main .\threadingconstructs.jl:214
     [4] #37#threadsfor_fun
       @ Main .\threadingconstructs.jl:181 [inlined]
     [5] (::Base.Threads.var"#1#2"{var"#37#threadsfor_fun#8"{var"#37#threadsfor_fun#7#9"{UnitRange{Int64}}}, Int64})()
       @ Base.Threads .\threadingconstructs.jl:153

...and 1 more exception.

Stacktrace:
[...]

PR #284, opened 2023-11-14
add support for using @showprogress on Threads.@threads for loops

Fro116:
This lets you use the @showprogress macro on parallel for loops. For example,
```
@showprogress Threads.@threads for i in 1:100
    sleep(0.1)
end
```

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -841,20 +841,13 @@ function showprogressdistributed(args...)
 end
 
 function showprogressthreads(args...)
-    if length(args) < 1
-        throw(ArgumentError("@showprogress Threads.@threads requires at least 1 argument"))
-    end
-
     progressargs = args[1:end-1]
     expr = args[end]
-    loop = expr
-    if loop.head == :macrocall && loop.args[1] == :(Threads.var"@threads")
-        loop = loop.args[end]
-    end
-
+    loop = expr.args[end]
+    r = loop.args[1].args[end]
     p = gensym()
     n = gensym()
-    r = loop.args[1].args[end]
+
     ex = quote
         $n = Int(round(length($(esc(r))) / Threads.nthreads()))
         global $p
@@ -869,6 +862,7 @@ function showprogressthreads(args...)
         end
     end
     push!(loop.args[end].args, update)
+
     ex
 end
 
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -844,26 +844,20 @@ function showprogressthreads(args...)
     progressargs = args[1:end-1]
     expr = args[end]
     loop = expr.args[end]
-    r = loop.args[1].args[end]
+    iters = loop.args[1].args[end]
+
     p = gensym()
-    n = gensym()
+    push!(loop.args[end].args, :(ProgressMeter.next!($p)))
 
-    ex = quote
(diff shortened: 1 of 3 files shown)


Fro116:
My bad, I never tested on older versions of Julia. The Threads.@threads had a different expression tree before v1.3.0-rc1 and so the new test I added was failing on older builds. 

I'll gate the Threads.@threads test behind a `@static if VERSION >= v"1.3.0-rc1"` statement, the same way we do for some Threads.@spawn tests in test_threads.jl.

Fro116:
Take three. I updated to fix the code coverage errors. 

The last run failed on the nightly ubuntu-latest build because the test case "test_threads.jl ProgressUnknown" errored out. I'm not sure why. That test should not touch any of the code in this pr. I can't repro it, and it didn't fail in the first run (https://github.com/timholy/ProgressMeter.jl/actions/runs/6859117251/job/18660114093?pr=284). Is the test flaky on master?

MarcMush:
Recently yes
Ref #281

Fro116:
Okay, afaict this passes everything except for the flaky test. Should be good to go, but let me know if there's anything else I should look into.

src/ProgressMeter.jl
```diff
@@ -840,6 +840,32 @@ function showprogressdistributed(args...)
...
+    progressargs = args[1:end-1]
+    expr = args[end]
+    loop = expr.args[end]
+    r = loop.args[1].args[end]
+    p = gensym()
+    n = gensym()
+
+    ex = quote
+        $n = Int(round(length($(esc(r))) / Threads.nthreads()))
+        global $p
```
MarcMush, line 853: is it necessary to make it global? this should probably be avoided

src/ProgressMeter.jl
```diff
@@ -840,6 +840,32 @@ function showprogressdistributed(args...)
...
+        $p = ProgressMeter.Progress($n; $(showprogress_process_args(progressargs)...))
+        $(esc(expr))
+        ProgressMeter.finish!($p)
+    end
+
+    update = quote
+        if Threads.threadid() == 1
+            ProgressMeter.next!(ProgressMeter.$p)
```
MarcMush, line 861: there is an example in the readme with a `Threads.@threads` that updates an all threads, can you try to make it work? The progressbar might be wrong for unevenly distributed tasks

test/test.jl
```diff
@@ -475,4 +475,19 @@ function testfunc22()
...
+        ProgressMeter.@showprogress dt=dt Threads.@threads for i in 1:n
+            if rand() < 0.7
+                sleep(tsleep)
+            end
+            result[i] = i ^ 2
+        end
+        @test sum(result) == sum(abs2.(1:n))
+    end
+    testfunc23(3000, 0.01, 0.001)
+end
```
MarcMush, line 493: can you rather put this in test_threads.jl ?

MarcMush (changes requested):
support for `@Threads.threads` is very welcome!

src/ProgressMeter.jl
MarcMush: you can probably use a `let`

src/ProgressMeter.jl
Fro116: Sure thing! I changed the code to update on every thread

MarcMush (approved):

MarcMush:
LGTM

PR #233, opened 2022-03-24
Add space after description if needed

IanButterworth:
Currently the `desc` is required to have a trailing space, otherwise you end up with
```
Custom message100%|█████████████████████| Time: 0:00:00
```

This just checks if there is a trailing space and adds one if not.
Intended to be the minimal way to fix this.

PR #285, opened 2023-11-14
add parameter for fractional bar length

blwh:
I find that the progress bar usually overflows when the ETA quickly converges from a large to a small value. Because of this, I want to make the bar shorter. However, since the TTY width is not immediately obvious to me, I would rather like to scale the length of the bar by some fraction. This PR adds this feature.

Summary of changes:
* Added the parameter `barlen_fraction` to the `Progress`-struct, which can take on any float. This means that the bar can be scaled negatively and larger than 1.

I might have added too much to the testing, and the argument might be too verbose.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -67,6 +67,7 @@ mutable struct Progress <: AbstractProgress
     printed::Bool              # true if we have issued at least one status update
     desc::String               # prefix to the percentage, e.g.  "Computing..."
     barlen::Union{Int,Nothing} # progress bar size (default is available terminal width)
+    barlen_fraction::Float64   # fraction of progress bar size
     barglyphs::BarGlyphs       # the characters to be used in the bar
     color::Symbol              # default to green
     output::IO                 # output stream into which the progress is written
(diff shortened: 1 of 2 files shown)


src/deprecated.jl
```diff
@@ -1,6 +1,6 @@
 @deprecate Progress(n::Integer, dt::Real, desc::AbstractString="Progress: ",
-    barlen=nothing, color::Symbol=:green, output::IO=stderr;
-    offset::Integer=0) Progress(n; dt=dt, desc=desc, barlen=barlen, color=color, output=output, offset=offset)
+    barlen=nothing, barlen_fraction::Float64=1.0, color::Symbol=:green, output::IO=stderr;
```
MarcMush, line 2: don't add it to the deprecated signature, we don't want to use these anymore. anyway it wouldn't be compatible with previous versions because it's not at the end

MarcMush:
for tests, add a few tests with the new feature but don't change old ones. That way we can check that the change is not breaking. All old tests should still pass (including the deprecated ones, which is not the case at the moment)

blwh:
I've added tests. As you noted, however, the performance test in `core.jl` fails. It seems to be due to the required truncation on line 209
```julia
    full_width = trunc(Int, (displaysize(output)[2]*width_fraction))
```
It's not consistent, as it only (mostly) fails on my machine. I ran the performance test `prog_perf(n)` on line 28 in `core.jl` and got on average that the code with the truncation is about 7% slower (0.144s vs 0.135s), which makes the test on line 51 in `core.jl` barely fail
```julia
    @test @elapsed(prog_perf(10^7)) < 9*@elapsed(noprog_perf(10^7))
```
I think the 9 is a bit arbitrary and that the potential slowdown of the truncation is not bad enough. If not, I do not have any other suggestions to improve the code.

MarcMush:
do you have a reproducible example of your original problem (the progressbar overflowing)?
I believe it might be a problem with #233 where a space is added without reducing the size of the progressbar

PR #289, opened 2023-12-11
Add gifs to the README to illustrate functionality

ericphanson:
I used ProgressMeter as a proving ground for my new package [Asciicast.jl](https://github.com/ericphanson/Asciicast.jl) which intends to make it easy and maintainable to have animated examples in github READMEs. (The package is not registered, but will be in a couple days).

This PR turns some of the README examples into Asciicast blocks, by adding the syntax `{cast = "true"}` to some of the julia code blocks. Then `Asciicast.cast_readme(ProgressMeter)` executes them (using the same sandboxes as Documenter) and saves the results as gifs (using [asciinema](https://asciinema.org/) & [agg](https://github.com/asciinema/agg)) to the `assets` directory, then uses pandoc to insert the image links back into the README.

This has the benefit of being easily maintainable, but the drawback that it uses pandoc's opinionated github-flavored markdown formatting, which causes a somewhat large diff here.

I am not sure if this is desired here, but since I was trying out Asciicast with ProgressMeter anyway, I figured I'd make a PR.


timholy:
This is pretty cool! I think it's a clear win in terms of usability for newbies. I guess one question is whether this will increase the package size: do the assets get downloaded when you install the package?

giordano:
> do the assets get downloaded when you install the package?

Yes, a snapshot of the entire repository is installed, but all the images in this PR are less than 100 kB each.

PR #291, opened 2024-01-18
tests: don't use single-argument `map`

nsajko:
Single-argument `map` is being removed from Julia after an analysis of registered packages found that the change doesn't break the functionality of any package. See JuliaLang/julia#35293 and JuliaLang/julia#52631

That said, the Julia change breaks your test suite, so here's a PR that fixes that.

### DIFF
--- a/test/test_map.jl
+++ b/test/test_map.jl
@@ -169,10 +169,8 @@ wp = WorkerPool(procs)
     function constfun()
         return 42
     end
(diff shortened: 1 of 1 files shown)


MarcMush:
`broadcast(f)` is also tested here, will it be also removed?

nsajko:
`broadcast(f)` is still supported.

PR #304, opened 2024-03-02
Drop Julia <1.6. Reduce code duplication by introducing ProgressCore

IanButterworth:
- Drops julia versions <1.6 (very old now, so not worth supporting given even things like `hasfield` aren't defined)
- Reduces code duplication .There's a fair bit of duplication between the three core Progress types. This puts the common elements into a `ProgressCore` struct to avoid duplication.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -45,6 +45,30 @@ function BarGlyphs(s::AbstractString)
     end
     return BarGlyphs(glyphs...)
 end
+const defaultglyphs = BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',)
+
+# Internal struct for holding common properties for progress meters
+Base.@kwdef mutable struct ProgressCore
+    lock::Threads.ReentrantLock = Threads.ReentrantLock()
+    dt::Real                    = Float64(0.1)
+    counter::Int                = 0
+    tinit::Float64              = time()
+    tsecond::Float64            = time()        # ignore the first loop given usually uncharacteristically slow
+    tlast::Float64              = time()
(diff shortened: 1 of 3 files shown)


src/ProgressMeter.jl
```diff
@@ -45,6 +45,30 @@ function BarGlyphs(s::AbstractString)
     end
     return BarGlyphs(glyphs...)
 end
+const defaultglyphs = BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',)
+
+# Internal struct for holding common properties for progress meters
+Base.@kwdef mutable struct ProgressCore
+    lock::Threads.ReentrantLock = Threads.ReentrantLock()
+    dt::Real                    = Float64(0.1)
```
MarcMush, line 53: maybe reorder, first the keywords that are meant to be changed by the user, then those who are internal only
`barlen` and `barglyph` are only for `Progress` and should probably go there

MarcMush:
that's great!

src/ProgressMeter.jl
```diff
@@ -57,46 +80,37 @@ the current task. Optionally you can disable the progress bar by setting
...
+    end
+end
+# forward common core properties to main types
+function Base.setproperty!(p::T, name::Symbol, value) where T<:AbstractProgress
+    if hasfield(T, name)
+        setfield!(p, name, value)
+    else
+        setproperty!(p.core, name, value)
+    end
+end
```
giordano, line 108: This is a breaking change, because it isn't handling automatic conversion of `value` to `T` like the default `setproperty!` method does: https://github.com/JuliaLang/julia/blob/a69aa30478ac9beda594235ff6f2e39549fdb752/base/Base.jl#L50-L54

PR #306, opened 2024-03-04
ProgressMeter v1.10.0

IanButterworth:
I'll leave this up for a day in case anyone thinks this should be a minor bump (all the commits looked like fixes to me)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,6 +38,7 @@ jobs:
       - uses: julia-actions/julia-runtest@v1
         env:
           JULIA_NUM_THREADS: ${{ matrix.nthreads }}
+        timeout-minutes: 20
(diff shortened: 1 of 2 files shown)


MarcMush:
I think #284 at least requires a minor bump

IanButterworth:
Hm good point, but that appears to be undocumented? no docstring, no mention in the readme?

IanButterworth:
Oh. I misread the `showprogressthreads` function as the API.  If this is making `@showprogress Threads.@threads for ...` work then IMO it seems more like a fix than a feature?

MarcMush:
I guess both can be justified
And I would argue that #296 is also a new feature

IanButterworth:
Ok. Let's do a minor bump then. Helps for messaging around the release.

IanButterworth:
https://github.com/JuliaRegistries/General/pull/102242

PR #307, opened 2024-03-04
update docs on `@showprogress` with threads

IanButterworth:
Follow on from https://github.com/timholy/ProgressMeter.jl/pull/284

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ end
 ```
 
 The core methods `Progress()`, `ProgressThresh()`, `ProgressUnknown()`, and their updaters
-are also thread-safe, so can be used with ``, `Threads.@spawn` etc.:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,7 +97,7 @@ end
 ```
 
 The core methods `Progress()`, `ProgressThresh()`, `ProgressUnknown()`, and their updaters
-are also thread-safe, so can be used with `Threads.@threads`, `Threads.@spawn` etc.:
+are also thread-safe, so can be used with ``, `Threads.@spawn` etc.:
```
MarcMush, line 100:
```suggestion
are also thread-safe, so can be used with `Threads.@threads`, `Threads.@spawn` etc.:
```

IanButterworth:
Let's merge this before #306

PR #309, opened 2024-04-03
fix breaking `setproperty!` introduced in #304

MarcMush:
fix https://github.com/timholy/ProgressMeter.jl/pull/304#discussion_r1549634629


MarcMush:
in case the case of https://github.com/timholy/ProgressMeter.jl/pull/304#discussion_r1549634629, `update!(p; max_steps)` should be preferred but we can still fix that

IanButterworth:
Thanks @MarcMush 

btw, I'm happy to review PRs here. This was self-merged a bit too quickly for me to review.

PR #315, opened 2024-06-25
Allow forcing the progress bar update

alyst:
The PR adds the `force=true/false`  keyword argument to `updateProgress!()` method.
When `force=true`, the progress bar is unconditionally updated regardless the too frequent bar updates check.

*The use case*: I am processing the big data file that I download chunk by chunk (one chunk contains thousands of data entries).
The progress bar reflects how many entries have been processed.
Each datum is processed within fraction of a second.
However, downloading the next chunk of data takes a while (minutes), during which time the progress bar freezes.
I use `showvalues=` to report the current status: either data processing or chunk download.
The problem is that in 99.9% cases the download-related progress bar update gets suppressed by the frequency filter (dt) -- so during the download the bar incorrectly reports that data entries are being processed.
So with this PR I can make sure that the bar would be updated just before the download by passing `force=true`.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -211,7 +211,8 @@ function updateProgress!(p::Progress; showvalues = (),
                          truncate_lines = false, valuecolor = :blue,
                          offset::Integer = p.offset, keep = (offset == 0),
                          desc::Union{Nothing,AbstractString} = nothing,
-                         ignore_predictor = false, color = p.color, max_steps = p.n)
+                         ignore_predictor = false, force::Bool = false,
+                         color = p.color, max_steps = p.n)
     !p.enabled && return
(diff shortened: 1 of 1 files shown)


MarcMush:
Does `dt=0` not work in this case?

alyst:
> Does dt=0 not work in this case?

It should definitely update the bar at each step, but that's not what I want:
I have ~10^8 fast steps and ~20 slow steps, and dt logic is perfectly fine for the small steps to avoid flooding the terminal.
I only want to make sure that the bar is updated for each slow step.

MarcMush:
indeed it makes sense
I thought `update!(p;dt=0)` would work but it doesn't

MarcMush (approved):

Issue #317, opened 2024-07-03
Julia 1.11: BoundsError: attempt to access MemoryRef{Int64} at index [1]


lmiq:
MWE:

1. Run with multi-threading (`julia -t auto`).

Code:

```julia
julia> using ProgressMeter

julia> function test(n)
           s = 0.0
           p = Progress(n)
           Threads.@threads for i in 1:n
               s += sum(exp(log(abs(sin(i)))) for i in 1:1000)
               next!(p)
           end
           return s
       end
test (generic function with 1 method)

julia> while true
           test(10000)
       end
ERROR: TaskFailedException

    nested task error: BoundsError: attempt to access MemoryRef{Int64} at index [1]
    Stacktrace:
      [1] GenericMemoryRef
        @ ./boot.jl:522 [inlined]
      [2] unsafe_copyto!
        @ ./genericmemory.jl:116 [inlined]
      [3] (::Base.var"#133#134"{Vector{Int64}, Int64, Int64, Int64, Int64, Int64, Memory{Int64}, MemoryRef{Int64}})()
        @ Base ./array.jl:1117
      [4] _growend!
        @ ./array.jl:1100 [inlined]
      [5] push!
        @ ./array.jl:1245 [inlined]
      [6] is_threading(p::Progress)
        @ ProgressMeter ~/.julia/packages/ProgressMeter/dMfiC/src/ProgressMeter.jl:446
      [7] lock_if_threading(f::ProgressMeter.var"#15#16"{Int64, @Kwargs{}, Progress}, p::Progress)
        @ ProgressMeter ~/.julia/packages/ProgressMeter/dMfiC/src/ProgressMeter.jl:452
      [8] #next!#14
        @ ~/.julia/packages/ProgressMeter/dMfiC/src/ProgressMeter.jl:471 [inlined]
      [9] next!
        @ ~/.julia/packages/ProgressMeter/dMfiC/src/ProgressMeter.jl:470 [inlined]
     [10] macro expansion
        @ ./REPL[11]:6 [inlined]
     [11] (::var"#35#threadsfor_fun#10"{var"#35#threadsfor_fun#8#11"{Progress, UnitRange{Int64}}})(tid::Int64; onethread::Bool)
        @ Main ./threadingconstructs.jl:252
     [12] #35#threadsfor_fun
        @ ./threadingconstructs.jl:219 [inlined]
[...]

Issue #232, opened 2022-03-21
is_threading is racy


dnadlinger:
In `is_threading`, `p.threads_used`, a `Vector` of thread ids, is accessed and updated:

https://github.com/timholy/ProgressMeter.jl/blob/5826a75b44078e7bce240f003885f02ee4d88d84/src/ProgressMeter.jl#L463-L470

I don't think `Vector` is thread-safe, though – wouldn't accesses `threads_used` also need some form of locking?

IanButterworth:
ah, true. It would be good to avoid a lock though, as the reason for this check is to avoid the need for a lock, given that locking got slower in recent julia versions

dnadlinger:
Which use case are you most concerned about, performance-wise (that isn't covered by the nthreads() fast path)? I don't see an obvious way to avoid locking altogether, other than avoiding the lock by using a lock-free counter to track work altogether.

Right now, I am personally only really interested in `@threads`. For this, something like
```julia
function is_threading()
  Threads.nthreads() == 1 && return false 

  if Threads.threadid() > 1
    # Called from another thread; so certainly in a @threads region.
    return true
  end

  # On the main thread, so just check if we are in a @threads region.
  return ccall(:jl_in_threaded_region, Cint, ()) != 0
end
```
This wouldn't work for `@spawn`/…, though.

IanButterworth:
It's quite expensive to lock each iteration on tight loops. If a user has `nthreads() > 1` and uses ProgressMeter on a non-threaded loop, I just felt the lock was worth avoiding.

It may be over-optimization though. Redesign proposals would definitely be welcomed by me!

dnadlinger:
I've honestly not used ProgressMeter.jl enough yet to be able to judge what the correct design tradeoffs are. One option to lower the overhead would be to use lock-free atomics instead, but then printing the updates could presumably only be done from the main thread to make sure the updates are ordered correctly. Using Threads.SpinLock might also be cheaper.

danielwe:
Workaround for anyone else who might be concerned about this:

```julia
p = Progress(n)
append!(p.threads_used, 1:Threads.nthreads())
# code using p
```

As for a proper solution, I think you can get away with only locking in the exceptional case, as follows:

```julia
function is_threading(p::AbstractProgress) 
    Threads.nthreads() == 1 && return false 
    length(p.threads_used) > 1 && return true 
    return lock(lk) do 
        if !in(Threads.threadid(), p.threads_used) 
[...]

PR #318, opened 2024-07-03
fix possible concurrency in is_threading

lmiq:
This fixes https://github.com/timholy/ProgressMeter.jl/issues/317

I think the issue is a concurrency that may occur in the `is_threading` function. 

I'm not sure if that function is necessary at all. The issue can be removed by always using the lock, and probably in serial runs the cost of that is irrelevant.

But maybe the bug reflects something deeper, so this might be the wrong fix.

(there is one test failing, but it is not related to this change)

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -441,25 +441,6 @@ end
 
 predicted_updates_per_dt_have_passed(p::AbstractProgress) = p.counter - p.prev_update_count >= p.check_iterations
 
-function is_threading(p::AbstractProgress)
-    Threads.nthreads() == 1 && return false
-    length(p.threads_used) > 1 && return true
-    if !in(Threads.threadid(), p.threads_used)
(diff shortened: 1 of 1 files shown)


MarcMush:
Ref #232

MarcMush:
As a workaround, you can imitate #298

lmiq:
> As a workaround, you can imitate #298

Ok, but this will break possibly many codes in Julia 1.11. A fix is now urgent. I'll just do the safe strategy and just always lock, and remove that. Also since tasks can migrate, I have the impression that using `threadid()` to detect parallelism is not a good idea anyway.

lmiq:
Closed because other PRs are dealing with this.

MarcMush:
I agree the current code is bad and needs to be fixed
Is there a performance hit? If it is non-negligible we can imagine a keyword like `thread-safe` that defaults to teue

PR #319, opened 2024-07-04
use Bool vector for threads_used

lmiq:
Now this is a possible fix for: https://github.com/timholy/ProgressMeter.jl/issues/317 and https://github.com/timholy/ProgressMeter.jl/issues/232

Instead of a vector of Ints, theading information is stored in the `ThreadingInfo` structure, which contains
two fields: `detected` and `threads_used`. The later plays the same role as the previous `threads_used`, but it
is a boolean array initialized as false for all threads. Accessing it is still racy, but in a safe way. The `detected` field stores the information that threading was detected, and then no more `threaid()` or updates to `threads_used` are necessary, maybe improving somewhat the performance.

I have not detected any performance difference relative to the previous approach, if any it was for better.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -65,6 +65,12 @@ function BarGlyphs(s::AbstractString)
 end
 const defaultglyphs = BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',)
 
+# Internal struct for holding threading information
+Base.@kwdef struct ThreadingInfo
+    detected::Bool = false
+    threads_used::Vector{Bool} = fill(false, Threads.nthreads())
+end
+
 # Internal struct for holding common properties and internals for progress meters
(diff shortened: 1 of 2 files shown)


src/ProgressMeter.jl
```diff
@@ -65,6 +65,12 @@ function BarGlyphs(s::AbstractString)
 end
 const defaultglyphs = BarGlyphs('|','█', Sys.iswindows() ? '█' : ['▏','▎','▍','▌','▋','▊','▉'],' ','|',)
 
+# Internal struct for holding threading information
+Base.@kwdef mutable struct ThreadingInfo
+    detected::Bool = false
+    threads_used::Vector{Bool} = fill(false, Threads.nthreads())
```
vchuravy, line 71: Threadid may return values bigger than nthreads.

When foreign threads are used (or even just with an interactive thread) iirc

lmiq: Oh, ok. Back to beginning.

PR #320, opened 2024-07-04
Locked threadid update

lmiq:
Another possible fix for: https://github.com/timholy/ProgressMeter.jl/issues/317 and https://github.com/timholy/ProgressMeter.jl/issues/232

In this case, the treading is detected by updating the `ThreadInfo` structure. This structure carries a bool, `detected`, that will cause the immediate return from `is_threading` once threading was detected. The field `threads_detected` will accumulate at most 2 different `threadid()`s, and is updated thread-safely with a lock. But this lock is only used until two different thread ids are detected, thus the cost of this should be minor.


lmiq:
Closed in favor of the simpler PR: https://github.com/timholy/ProgressMeter.jl/pull/321

Issue #253, opened 2022-10-31
Progress output is mangled when using `async`


arnaudh:
```
using ProgressMeter
progress = Progress(10);
asyncmap(1:10) do i
   sleep(i*0.1)
   next!(progress)
end
```

Output:
```
Progress:  20%|███████████▍                                             |  ETA: 0:00:04Progress:  30%|█████████████████▏                                       |  ETA: 0:00:03Progress:  40%|██████████████████████▊                                  |  ETA: 0:00:02Progress:  50%|████████████████████████████▌                            |  ETA: 0:00:01Progress:  60%|██████████████████████████████████▎                      |  ETA: 0:00:01Progress:  70%|███████████████████████████████████████▉                 |  ETA: 0:00:01Progress:  80%|███████████████████████████████████████Progress: 100%|█████████████████████████████████████████████████████████| Time: 0:00:01
```

Seems like the line doesn't get properly cleared/overriden when multiple calls to `next!(progress)` happen concurrently.

Note if I bump the sleep to a higher value (`sleep(i)`), the issue goes away:
```
Progress: 100%|█████████████████████████████████████████████████████████| Time: 0:00:10
```

- Julia v1.8
- ProgressMeter v1.7.2

arnaudh:
Had a quick look at the code and documentation, seems like `@showprogress` can work with `asyncmap`, and a quick test shows the issue is not present in that case:
```
@showprogress asyncmap(1:10) do i
   sleep(i*0.1)
end
```
Output:
```
Progress: 100%|█████████████████████████████████████████████████████████| Time: 0:00:02
```

Sounds like this is the way to go.

Note I assumed my example (using `ayncmap` and `next!`) would be supported because of the note on thread-safety in the README (with the example using `@threads` and `next!`). Might be worth mentioning that `asyncmap` is only supported via  `@showprogress`.

MarcMush:
The problem appears when two parallel tasks try to erase and write at the same time

Using `next!` with a `ParallelProgress` will be possible with #157

MarcMush:
this should now work by forcing the lock (`Progress(10; safe_lock=true)`) (#322)

PR #321, opened 2024-07-04
simple lock for threads_used update

lmiq:
Simple fix for https://github.com/timholy/ProgressMeter.jl/issues/317 and https://github.com/timholy/ProgressMeter.jl/issues/232

This PR has the same effect as https://github.com/timholy/ProgressMeter.jl/pull/320 but is a simpler modification of the original code. I just added a lock to update `threads_used`. The update of `threads_used` occurs only twice at most, so this lock is very likely not important for performance.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -77,11 +77,11 @@ Base.@kwdef mutable struct ProgressCore
     # internals
     check_iterations::Int       = 1             # number of iterations to check time for
     counter::Int                = 0             # current iteration
-    lock::Threads.ReentrantLock = Threads.ReentrantLock() # lock used when threading detected
+    lock::Threads.ReentrantLock = Threads.ReentrantLock()  # lock used when threading detected
     numprintedvalues::Int       = 0             # num values printed below progress in last iteration
     prev_update_count::Int      = 1             # counter at last update
     printed::Bool               = false         # true if we have issued at least one status update
-    threads_used_lock::ReentrantLock = ReentrantLock() # lock for threads_used update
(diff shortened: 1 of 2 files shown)


lmiq:
This function (which is used in a test which is failing in both master and this updated version) should measure the cost of this lock. I don't see any measurable difference in performance among the versions:

With this PR:

```julia
ulia> function prog_perf(n)
           prog = Progress(n)
           x = 0.0
           for i in 1:n
               x += rand()
               next!(prog)
           end
           return x
       end
prog_perf (generic function with 1 method)

julia> @elapsed(prog_perf(10^7))
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
4.229378385

julia> @elapsed(prog_perf(10^7))
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
4.185957772
```

With current stable release:

```julia
julia> @elapsed(prog_perf(10^7))
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
4.125580045

julia> @elapsed(prog_perf(10^7))
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
4.197386729
```

MarcMush:
can you test the performance by always locking? I would be in favor of the `no_lock` keyword. It would also fix `@async` (#253), which cannot be detected

lmiq:
> can you test the performance by always locking? I would be in favor of the `no_lock` keyword. It would also fix `@async` (#253), which cannot be detected

Added the option here: https://github.com/timholy/ProgressMeter.jl/pull/322

with some performance comparison (I think we should always lock for good :-) )

PR #322, opened 2024-07-04
lock by default - add safe_lock option for control

lmiq:
This is another possible fix for: https://github.com/timholy/ProgressMeter.jl/issues/317 and https://github.com/timholy/ProgressMeter.jl/issues/232

Here we remove the automatic identification of threading, and always use a lock by default. An option `safe_lock` is added, which is by default `true` and must be set to `false` to disable the lock.

Relative to the current stable version I don't see any measurable performance difference:

Current stable:

```julia
julia> function prog_perf(n)
           prog = Progress(n)
           x = 0.0
           for i in 1:n
               x += rand()
               next!(prog)
           end
           return x
       end
prog_perf (generic function with 1 method)

julia> @time prog_perf(10^7)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:05
  5.938973 seconds (122.22 M allocations: 1.889 GiB, 2.62% gc time, 28.52% compilation time)
4.999510125125184e6

julia> @time prog_perf(10^7)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:03
  3.953561 seconds (119.76 M allocations: 1.785 GiB, 2.68% gc time)
4.999360964692622e6

julia> @time prog_perf(10^7)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
  4.044011 seconds (119.74 M allocations: 1.784 GiB, 2.60% gc time)
5.0003511864208765e6
```

With this PR:

```julia
julia> function prog_perf(n; safe_lock=true)
           prog = Progress(n; safe_lock)
           x = 0.0
           for i in 1:n
               x += rand()
               next!(prog)
           end
           return x
       end
prog_perf (generic function with 1 method)

julia> @time prog_perf(10^7)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:04
  4.068567 seconds (119.76 M allocations: 1.785 GiB, 4.76% gc time)
5.002101485254395e6

julia> @time prog_perf(10^7)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:03
  3.992197 seconds (119.74 M allocations: 1.784 GiB, 3.91% gc time)
4.999887783558271e6

julia> @time prog_perf(10^7; safe_lock=false)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:03
  3.704703 seconds (119.74 M allocations: 1.784 GiB, 4.53% gc time)
4.998432809492471e6

julia> @time prog_perf(10^7; safe_lock=false)
Progress: 100%|█████████████████████████████████████████████████████| Time: 0:00:03
  3.749583 seconds (119.73 M allocations: 1.784 GiB, 4.10% gc time)
5.000024621577769e6
```

There is an apparent small but consistent advantage when `safe_lock=false`, which is expected.

### DIFF
--- a/test/core.jl
+++ b/test/core.jl
@@ -92,10 +92,7 @@ function simple_sum(n; safe_lock = true)
     return s
 end
 p = Progress(10)
-@test p.safe_lock == true
+@test p.safe_lock == Threads.nthreads() > 1
 p = Progress(10; safe_lock = false)
 @test p.safe_lock == false
-@test simple_sum(10) ≈ simple_sum(10; safe_lock = false)
-
-
-
+@test simple_sum(10; safe_lock = false) ≈ simple_sum(10; safe_lock = false)
--- a/test/core.jl
+++ b/test/core.jl
@@ -92,7 +92,7 @@ function simple_sum(n; safe_lock = true)
     return s
 end
 p = Progress(10)
-@test p.safe_lock == Threads.nthreads() > 1
+@test p.safe_lock == (Threads.nthreads() > 1)
 p = Progress(10; safe_lock = false)
 @test p.safe_lock == false
 @test simple_sum(10; safe_lock = false) ≈ simple_sum(10; safe_lock = false)
--- a/test/core.jl
+++ b/test/core.jl
@@ -95,4 +95,4 @@ p = Progress(10)
 @test p.safe_lock == (Threads.nthreads() > 1)
 p = Progress(10; safe_lock = false)
 @test p.safe_lock == false
-@test simple_sum(10; safe_lock = false) ≈ simple_sum(10; safe_lock = false)
+@test simple_sum(10; safe_lock = true) ≈ simple_sum(10; safe_lock = false)
--- a/test/core.jl
+++ b/test/core.jl
@@ -81,22 +81,21 @@ prog.n = UInt128(20) # in Progress
 prog.offset = Int8(5) # in ProgressCore
 @test prog.offset == 5
 
-# Test disable_lock option, initialization and
-# execution. 
-function simple_sum(n; disable_lock = false)
-    p = Progress(n; disable_lock)
+# Test safe_lock option, initialization and execution. 
+function simple_sum(n; safe_lock = true)
+    p = Progress(n; safe_lock)
     s = 0.0
     for i in 1:n
-        s += sin(i)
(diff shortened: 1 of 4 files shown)


MarcMush (approved):
looks good

lmiq:
Shoudn´t we release one of these fixes? None is breaking, and while the ˋsafe_lockˋ option is not documented, it can be reverted. If a better solution if proposed later, it can be implemented anyway.

lmiq:
In https://github.com/timholy/ProgressMeter.jl/issues/232#issuecomment-1685158237 @IanButterworth mentions that the use of the progress meter slows down 9x the tight loop. I actually see a much greater slowdown with the released version, but my tests with this version, for which the lock can be removed, show that the slowdown is not really related to the lock.

lmiq:
The main issue I see here is that the comparison with **not** using the progress meter is so great, that it makes me wander if this discussion about performance is really relevant. This is what I get here with the functions used in the test suite for this:

```
julia> @time noprog_perf(10^7)
  0.014616 seconds
5.001148339582725e6

julia> @time prog_perf(10^7)
Progress: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:05
  5.515208 seconds (119.79 M allocations: 1.785 GiB, 0.55% gc time)
5.000329702123166e6

julia>
```

This is with the current released version. Does it make sense to discuss the performance differences of using or not a lock in this context? Seems to me that using it for a tight loop simply does not make sense with the current status of things.

lmiq:
Tests fixed.

test/core.jl
```diff
@@ -80,3 +80,19 @@ prog.n = UInt128(20) # in Progress
...
+        s += sin(i)^2
+        next!(p)
+    end
+    return s
+end
+p = Progress(10)
+@test p.safe_lock == (Threads.nthreads() > 1)
+p = Progress(10; safe_lock = false)
+@test p.safe_lock == false
+@test simple_sum(10; safe_lock = false) ≈ simple_sum(10; safe_lock = false)
```
MarcMush, line 98: did you mean to compare `true` with `false`?

lmiq: Yes, sorry. fixed.

PR #325, opened 2024-07-18
use getfield for performance and add performance tests

### LINKED ISSUE #323: Terrible performance with Julia 1.10+ProgressMeter 1.10
```julia
using ProgressMeter

function prog_perf(n)
    prog = Progress(n, enabled=false)
    x = 0.0
    for i in 1:n
        x += rand()
        next!(prog)
    end
    return x
end

function noprog_perf(n)
    x = 0.0
    for i in 1:n
        x += rand()
    end
    return x
end
```
time on windows:
noprog: ~0.01 sec
julia 1.10, ProgressMeter 1.10: 8.0 sec
julia 1.9, ProgressMeter 1.10 : 0.17 sec
julia 1.11-rc1, ProgressMeter 1.10 : 2.28 sec
julia 1.10, ProgressMeter 1.9: 0.11 sec

similar times on Linux

profiling seems to inidicate a lot of time in `getproperty` and `setproperty!`, so it might be because of `ProgressCore`, but it doesn't happen in julia 1.9 so it's weird

MarcMush:
fix #323

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -206,14 +206,18 @@ function calc_check_iterations(p, t)
     return round(Int, clamp(iterations_per_dt, 1, p.check_iterations * 10))
 end
 
+function updateProgress!(p::AbstractProgress; options...)
+    !p.enabled && return nothing
+    _updateProgress!(p; options...)
+end
+
 # update progress display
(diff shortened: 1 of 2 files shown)


IanButterworth:
Great!

PR #328, opened 2024-07-19
Return of the threaded detection

MarcMush:
with #325, the difference in performance with the `safe_lock` is much more noticeable compared to #322 (x4-5 on my machine) 

this stores only the starting thread instead of a vector of all used threads (which is basically the same)

I can't see a repeatable difference between disabling the lock and letting in detect but both are still possible with this

Performance results from `runtest.jl`: (these numbers vary quite a lot between each run)
```
Performance results:
without progress:      1.23 ns/it
with no lock:          9.23 ns/it
with no printing:      8.85 ns/it
with disabled:         2.60 ns/it
with lock:            41.06 ns/it
with automatic lock:   9.02 ns/it
with lock, disabled:  32.18 ns/it
with force:            0.20 ms/it

Threaded performance results: (4 threads)
without progress:     64.75 ns/it
with automatic lock:   0.23 μs/it
with forced lock:      0.23 μs/it
with no printing:      0.24 μs/it
with disabled:         0.20 μs/it
with force:            0.25 ms/it
```

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -81,7 +81,8 @@ Base.@kwdef mutable struct ProgressCore
     numprintedvalues::Int       = 0             # num values printed below progress in last iteration
     prev_update_count::Int      = 1             # counter at last update
     printed::Bool               = false         # true if we have issued at least one status update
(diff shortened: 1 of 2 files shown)


MarcMush:
I'm open to other suggestions

Issue #333, opened 2024-09-29
Redesign using `Threads.@spawn`


timholy:
This package was started long before Julia had threading. Ideally, the decision about whether to print a progress update should be made in a separate thread, so that if the first step is long-running you at least see something within the delay time.

PR #334, opened 2024-11-13
feat: add thread-safe progress meter

younes-io:


MarcMush:
it looks ok but is there a problem you came up across? Does this test something that wasn't tested before?

younes-io:
> it looks ok but is there a problem you came up across? Does this test something that wasn't tested before?

I added this to test how threads behave when they're all trying to update the same thing at once, but with different timing patterns. This can catch subtle issues that simpler tests might miss.

MarcMush:
we've had problems with threading (#232 and linked discussions) and there can be more done to improve multi-threading (#333, #157)

I wrote this to try to catch unsafe threading
https://github.com/timholy/ProgressMeter.jl/blob/cd0414fa75b51fae39c16436021002cfee21bb88/test/core.jl#L174-L184

I think no sleep is a worst-case, and sleep yields (cpu can do other stuff) which is not a realistic workload either

I would have no problem adding these tests but i don't see what they add. If you have more insight on threaded stuff please say, I'm pretty new to it

younes-io:
@MarcMush you're right, those tests were just simulating artificial computation.. I wasn't aware of [this issue](https://github.com/timholy/ProgressMeter.jl/issues/333), so I pushed a commit to implement @timholy's recommendation using `Threads.@spawn`
Obviously, this is just a start.. Please, let me know your thoughts

PR #337, opened 2025-02-23
pretty print Progress

xgdgsc:
I find the defaultglyphs printed inline in VSCode too long and weird.

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -2,6 +2,7 @@ module ProgressMeter
 
 using Printf: @sprintf
 using Distributed
+import Base.show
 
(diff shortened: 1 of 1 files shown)


MarcMush:
from https://docs.julialang.org/en/v1/manual/types/#man-custom-pretty-printing:

> As a rule of thumb, the single-line show method should print a valid Julia expression for creating the shown object. 

so we should print every non-default parameter I guess, nearly all of them can be changed

it should also include `ProgressUnknown` and `ProgressThresh`

otherwise, I believe you can just add a semi colon at the end of the line to hide the output

xgdgsc:
I don' t think a daily utility 99% users would only change or care about the `n` should follow that rule or see anything inside ProgressCore . I' ve never used `ProgressUnknown ` or `ProgressThresh` so I wouldn' t change that.

PR #341, opened 2025-04-02
Fix printing of multiline values, and allow colors in showed values

MasonProtter:
Closes #224. That solution only worked if the string was padded to the display width, but this solution should correctly be able to deal with multiline strings.

Here's a demo:
```julia
let n = 20
    xs = Float64[]
    p = Progress(n)
    for iter = 1:n
        append!(xs, rand())
        sleep(0.5)
        plt = lineplot(xs)
        ProgressMeter.next!(p; showvalues = [(:UnicodePlot, plt)])
    end
end
```
[Screencast_20250402_171908.webm](https://github.com/user-attachments/assets/ba31eae8-4742-4627-b68b-2dc6a1c1fb0e)

PR #342, opened 2025-04-03
v1.10.3

IanButterworth:
@MasonProtter I think this can be a patch release. Would you mind reviewing commits since the last release and seeing if you agree?


MasonProtter:
Looks to me like
* README change (#329)
* Printing alignment change (#330)
* Internal changes to how the thread safety is handled (#328)
* Changes to the test suite (#339)
* My PR, a bugfix/enhancement for how values are displayed (#341)

So I see no API changes or anything else that could break stuff.

PR #345, opened 2025-04-04
Retry Fix printing of multiline values, and allow colors in showed values

MasonProtter:
This is a continuation of #341 after it was reverted because it errored when encountering an `AnnotatedString`.

Seems to be working with `AnnotatedString`.
```julia
julia> let n = 20
           xs = Float64[]
           p = Progress(n)
           for iter = 1:n
               append!(xs, rand())
               sleep(0.5)
               plt = lineplot(xs)
               str = Base.AnnotatedString(" julia", [(2:6, :face, rand((:magenta, :red, :blue, :green)))])
               ProgressMeter.next!(p; showvalues = [(:UnicodePlot, plt), ("An annoted string", str2)])
           end
       end
```
[Screencast_20250405_002203.webm](https://github.com/user-attachments/assets/ff36211a-3219-474f-952e-543511bd77ea)

The downside is that styled strings will get a `"` wrapped around them as you can see in the video, so if anyone has any better suggestions for how to deal with this generically, I'd be open to input. cc @MilesCranmer

### DIFF
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -580,13 +580,10 @@ function printvalues!(p::AbstractProgress, showvalues; color = :normal, truncate
     p.numprintedvalues = 0
 
     for (name, value) in showvalues
-        string_value = "\e[0m" * if value isa AbstractString
+        string_value = "\e[0m" * if value isa Union{String, SubString{String}}
             value
         else
-            let io = PipeBuffer()
-                show(IOContext(io, :color => true), value)
-                read(io, String)
-            end
+            repr("text/plain", value; context=IOContext(PipeBuffer(), :color => true))
         end
         if countlines(IOBuffer(string_value)) > 1
             # Multiline objects should go on their own line so their
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -580,7 +580,10 @@ function printvalues!(p::AbstractProgress, showvalues; color = :normal, truncate
     p.numprintedvalues = 0
 
     for (name, value) in showvalues
-        string_value = string(value, color=true)
+        string_value = let io = PipeBuffer()
+            show(IOContext(io, :color => true), value)
+            "\e[0m" * read(io, String)
+        end
         if countlines(IOBuffer(string_value)) > 1
             # Multiline objects should go on their own line so their
             # alignment doesn't get messed up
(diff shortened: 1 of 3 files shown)


MilesCranmer:
Probably worth adding some tests; also see the code coverage

src/ProgressMeter.jl
```diff
@@ -580,11 +580,21 @@ function printvalues!(p::AbstractProgress, showvalues; color = :normal, truncate
...
+        end
+        if countlines(IOBuffer(string_value)) > 1
+            # Multiline objects should go on their own line so their
+            # alignment doesn't get messed up
+            string_value = "\n" * string_value
+        end
+        msg = "\n  " * lpad(string(name) * ": ", maxwidth+2+1) * string_value
         max_len = (displaysize(p.output)::Tuple{Int,Int})[2]
         # I don't understand why the minus 1 is necessary here, but empircally
         # it is needed.
```
MilesCranmer, line 596: Does this "empirically it is needed" change at all with this modified calculation? Is there a test for this that verifies it is/is not needed?

src/ProgressMeter.jl
```diff
@@ -580,11 +580,21 @@ function printvalues!(p::AbstractProgress, showvalues; color = :normal, truncate
     p.numprintedvalues = 0
 
     for (name, value) in showvalues
-        msg = "\n  " * lpad(string(name) * ": ", maxwidth+2+1) * string(value)
+        string_value = "\e[0m" * if value isa Union{String, SubString{String}}
+            value
+        else
+            repr("text/plain", value; context=IOContext(PipeBuffer(), :color => true))
```
MilesCranmer, line 586: Is this repr the cause of the explicit quotes showing up?

MasonProtter: Yes. Trying to `show` it like in the previous commit causes it to show up as `AnnotatedString("    julia", ...)`

src/ProgressMeter.jl
MasonProtter: That's an old comment from before my PR. I did try some tests to change that value, and it causes the number of printed lines to shrink or expand each iteration if you change it.

MasonProtter:
Yeah I'll throw some tests together soon, I just wanted to put the PR up for now to see if  people have suggestions for how to deal with these annotated strings

src/ProgressMeter.jl
MilesCranmer: Yeah I assumed it was old, I was just wondering if it was no longer relevant since it’s now a different method for counting lines

src/ProgressMeter.jl
MarcMush: maybe with `sprint`? https://docs.julialang.org/en/v1/stdlib/StyledStrings/#Applying-faces-to-a-AnnotatedString

PR #353, opened 2025-07-28
Set `IJulia.stdio_bytes` in a forward-compatible way

JamesWrigley:
The property will change to a field of a struct in an upcoming release: https://github.com/JuliaLang/IJulia.jl/pull/1145

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -7,7 +7,7 @@ Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 
 [compat]
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
--- a/src/ProgressMeter.jl
+++ b/src/ProgressMeter.jl
@@ -617,7 +617,12 @@ function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
     print(io, "\r")
     printstyled(io, s; color=color)
     if isdefined(Main, :IJulia)
-        Main.IJulia.stdio_bytes[] = 0 # issue #76: circumvent IJulia I/O throttling
+        # issue #76: circumvent IJulia I/O throttling
+        if pkgversion(Main.IJulia) < v"1.30"
+            Main.IJulia.stdio_bytes[] = 0
+        else
+            Main.IJulia.reset_stdio_count()
+        end
     elseif isdefined(Main, :ESS) || isdefined(Main, :Atom)
     else
         print(io, "\u1b[K")     # clear the rest of the line


halleysfifthinc:
This might not be the right time/place to suggest, but this kind of interaction and issue would be better solved by adding an IJulia extension, which would solve the compat aspect of this.

JamesWrigley:
That's true, this is all I have time for ATM but yeah ideally both IJulia and Distributed support would be in extensions.

Project.toml
```diff
@@ -7,7 +7,7 @@ Distributed = "8ba89e20-285c-5b6f-9357-94700520ee1b"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 
 [compat]
-julia = "1.6"
+julia = "1.10"
```
MarcMush, line 10: is bumping julia version necessary?

MarcMush: nvm it's for `pkgversion`

MarcMush:
ProgressMeter.jl has a lot of these, which look difficult to move into extensions without type-piracy:

```julia
function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
    print(io, "\r")
    printstyled(io, s; color=color)
    if isdefined(Main, :IJulia)
        Main.IJulia.stdio_bytes[] = 0 # issue #76: circumvent IJulia I/O throttling
    elseif isdefined(Main, :ESS) || isdefined(Main, :Atom)
    else
        print(io, "\u1b[K")     # clear the rest of the line
    end
end
```

an extension for Distributed.jl looks more doable

JamesWrigley:
I've changed this to use a new public function that will be used: https://julialang.github.io/IJulia.jl/dev/library/public/#IJulia.reset_stdio_count

I think it's ready to be merged now that the IJulia PR has been merged.

MarcMush:
should we first add a version of ProgressMeter that has a `[weakdep]` to limit to the old ijulia version?

JamesWrigley:
We could, but I don't think that's necessary with the current code? Or are you saying that we make one release with the current code and a weakdep for IJulia 1.29, and then another version with the new code I added and a weakdep on IJulia 1.30?

MarcMush:
yes that's what I meant
It seems cleaner but it might not matter, we can merge like that I guess

JamesWrigley:
Come to think of it I'm actually not sure that would work in a lot of cases because of how weird IJulia is :sweat_smile: Many users install it in their global environment instead of in notebook environments, so someone could be in a situation where the global environment has an old IJulia version and new ProgressMeter version. AFAIK compat bounds aren't always respected with stacked environments so that would cause an error if ProgressMeter only supported the newer IJulia version (and vice-versa).

JamesWrigley:
(minor change in f3a99ae to make the code use `Main.IJulia` for consistency)

src/ProgressMeter.jl
```diff
@@ -617,7 +617,12 @@ function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
     print(io, "\r")
     printstyled(io, s; color=color)
     if isdefined(Main, :IJulia)
-        Main.IJulia.stdio_bytes[] = 0 # issue #76: circumvent IJulia I/O throttling
+        # issue #76: circumvent IJulia I/O throttling
+        if pkgversion(Main.IJulia) < v"1.30"
```
stevengj, line 621: Couldn't this have used `isdefined(Main.IJulia, :reset_stdio_count)` instead?

JamesWrigley: Yep that would work too, I just prefer this style.

PR #354, opened 2025-08-23
v1.11.0

MarcMush:
or should this be 1.11.0?


timholy:
Since we're changing the minimal supported Julia version, this should be 1.11.0. After that change, this is approved to merge.

PR #356, opened 2026-01-05
Add `showelapsed` and `showeta` options to `Progress`

### LINKED ISSUE #287: Feature request: Add option to show elapsed time since start
Sometimes one wants the progress bar to show how long the task is already running.
Therefore, I suggest adding an option like `showspeed=true` but here `showelapsed=true`
which also shows the time since the loop started.

I guess the implementation is easy and straightforward.

Would love to see this in the future!

lgoettgens:
Resolves https://github.com/timholy/ProgressMeter.jl/issues/287.

This adds the possibility to set `showelapsed=true` for a `Progress` object to additionally show the total elapsed time. With `showeta=false`, one can further hide the current eta output.
To make this work, I had to refactor `tty_width` quite a bit, but I think the new comments there make it really clear what's happening.

I am happy to hear your feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,37 @@ next!(p; showvalues = generate_showvalues(iter, x))
 end
 ```
 
+### Showing elapsed time
+
+When using a `Progress`, it might be useful to show the elapsed time
+in addition to or instead of the estimated time.
+You can include elapsed time in your progress meter
+by setting the optional keyword argument `showelapsed=true`.
+If you want to additionally disable the estimated time,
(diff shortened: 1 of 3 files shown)


lgoettgens:
@timholy would you mind having a look at this? Thanks!

lgoettgens:
Gentle bump

MarcMush:
I think these are absolutely options that ProgressMeter.jl should have but I'm hesitant to add new printing options before tidying printing with for example something like #357

PR #366, opened 2026-09-11
Move Distributed support to a package extension

### LINKED ISSUE #332: Feature request: move Distributed to package extension
Would there be interest in moving the `showprogressdistributed` function to a package extension on Distributed?  It could shave ~30 ms off load times for any package that uses ProgressMeter.

I could make a PR fairly easily, but the disadvantage is that it locks new versions of the package to support a minimum Julia version of v1.9.

What do folks here think?

asinghvi17:
Closes #332.

Distributed becomes a weak dependency. Its support lives in `ProgressMeterDistributedExt`, which loads with `using Distributed`, and `using ProgressMeter` drops from 15 ms to 4 ms on Julia 1.10.

## Changes

- ProgressMeter declares `showprogressdistributed` and `progress_pmap` with their docstrings; the extension adds the methods. Calling either before Distributed loads throws a `MethodError` whose hint names `using Distributed`.
- The `@distributed` expansion interpolates `Distributed.@distributed` and `RemoteChannel`, since unescaped names in a `@showprogress` expansion resolve in ProgressMeter.
- The `ncalls` methods for `pmap` move to the extension.
- `ProgressMeter.progress_channel(mapfun, bufflen)` creates the channel `progress_map` reports through. The fallback is a local `Channel`; the extension's `::Function` method returns a `RemoteChannel`, so with Distributed loaded every `mapfun` reports as on master, including user-defined ones that run `f` on workers. A `mapfun` known to run on the main process can opt into the local channel with a `progress_channel(::typeof(mapfun), bufflen)` method.

## Behaviour

With Distributed loaded anywhere in the session, behaviour matches master. Without it, `@showprogress @distributed` and `progress_pmap` throw the hinted `MethodError`. On master they ran serially, since a session without Distributed has no workers. A minor version bump at release should cover this.

## Tests

- A "Without Distributed" testset runs first and checks the absent extension, the local channel in `progress_map`, and both hinted errors.
- `test_map.jl` checks that a user-defined `mapfun` running `f` on workers advances the counter.
- The full suite passes with `CI=true` on Julia 1.10.12 (1 thread) and 1.12.7 (2 threads).


src/ProgressMeter.jl
```diff
@@ -894,7 +825,9 @@ function showprogress(args...)
         macroname = expr.args[1]
 
         if macroname in (Symbol("@distributed"), :(Distributed.var"@distributed"))
-            return showprogressdistributed(args...)
+            ext = Base.get_extension(@__MODULE__, :ProgressMeterDistributedExt)
+            ext === nothing && throw(ArgumentError("@showprogress @distributed requires `using Distributed`"))
+            return ext.showprogressdistributed(args...)
```
asinghvi17: This is not a good pattern.  Instead, there should be a name here, `function somename end`, which gets defined in the extension.  You can even add an error hint or so.

asinghvi17:
Looks like we can get rid of the 1.11-nightly CI?  And not sure what's going on with the regular nightly.

asinghvi17:
oh, looks like nightly is https://github.com/JuliaLang/julia/issues/62197

MarcMush:
looks good to me, it doesn't seem to have any breaking changes (I guess technically but it would be very weird to use distributed features without importing it explicitly)

I've removed 1.11-nightly and rewrote the tests to avoid `@test_deprecated begin ... end`

for the tests, I think it would be better to do most of the existing tests without Distributed, to check that all functionnalities actually work without Distributed, and add a new file for `test_distributed.jl`, where all tests using Distributed are grouped

`runtests.jl` would look like this:

```julia
using ProgressMeter
using Test

if get(ENV, "CI", "false") == "true"
    using InteractiveUtils
    display(versioninfo())   # among other things, this shows the number of threads
end

@testset "Core" begin
    include("core.jl")
    include("test.jl")
end
@testset "Show Values" begin
    include("test_showvalues.jl")
end
@testset "Mapping" begin
    include("test_map.jl")
end
@testset "Float" begin
    include("test_float.jl")
end
@testset "Threading" begin
    include("test_threads.jl")
end
@testset "Deprecated" begin
    include("deprecated.jl")
end

using Distributed

@testset "DistributedExt" begin
    include("test_distributed.jl")
end
```

MarcMush:
It all looks good to me

I'll wait a few days before merging in case other maintainers want to have a look

asinghvi17:
thanks for the review!

bjarthur:
i had claude build a small MWE and verified that this branch, unlike v1.11.0, can be compiled with `juliac --trim`!!!

      MWE: using ProgressMeter alone fails juliac --trim=safe on v1.11.0, passes on #366

      PMTrim/Project.toml
      name = "PMTrim"
      uuid = "5c2e9f1a-7b3d-4a6e-8f0c-1d2e3f4a5b6c"
      version = "0.1.0"

      [deps]
      ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"

      To build against #366, add:
      [sources]
      ProgressMeter = {url = "https://github.com/asinghvi17/ProgressMeter.jl", rev =
      "distributed-ext"}

      PMTrim/src/PMTrim.jl
      module PMTrim
      using ProgressMeter
      function (@main)(args)
          println(Core.stdout, "ProgressMeter loaded")
          return 0
      end
      end

      cd PMTrim
      juliac --output-exe pmtrim --trim=safe --bundle ../build .
      ../build/bin/pmtrim

      Julia 1.13.0, JuliaC 0.3.10, Linux aarch64:

      ┌────────────────────┬─────────────────┐
      │   ProgressMeter    │ verifier errors │
      ├────────────────────┼─────────────────┤
      │ registered v1.11.0 │ 9               │
      ├────────────────────┼─────────────────┤
      │ #366 head 78f149b  │ 0               │
      └────────────────────┴─────────────────┘

      All nine errors on v1.11.0 are in a Base closure with no ProgressMeter frame,
      Base.var"#errormonitor##0#errormonitor##1"{Task}, on IOContext, write, flush, showerror,
      and printstyled against Base.stderr::IO. The first:

      Verifier error #1: unresolved call from statement Base.IOContext(%new()::IOBuffer,
      Base.stderr::IO)::IOContext{IOBuffer}

bjarthur:
note that i have two more PRs in the works to really get ProgressMeter to work with juliac.  one makes the IJulia dependency an extension, and the second makes a whole bunch of types trim safe.  stay tuned.

@leostenzel

bjarthur:
for those that want to test juliac, i have merged PRs 366, 367, and 368 here: https://github.com/bjarthur/ProgressMeter.jl/tree/bja/integration

PR #367, opened 2026-09-15
Make progress meters compile under juliac --trim=safe

bjarthur:
A trimmed binary that updates a meter currently fails the trim verifier with
dozens of errors, all from dynamic dispatch inside ProgressMeter:

- `ProgressCore.output::IO` makes every `print`, `flush`, `displaysize`, and
  `printover` on the output a dynamic call.
- `dt::Real` makes `tlast + dt` and `calc_check_iterations` dynamic, so
  `check_iterations` is assigned an `Any`.
- `"\n" ^ n` and `repeat` in `barstring` are unresolved invokes.
- `printstyled` cannot be compiled: `with_output_color` calls an untyped
  `f::Function`.
- `compute_front` reads the mutable `BarGlyphs.front` twice, so its return
  type is inferred as `Union{Char, Vector{Char}}` and `print` compiles a
  `show_vector` path.

## Changes

- `ProgressCore`, `Progress`, `ProgressThresh` and `ProgressUnknown` are
  parametrized on the output type `O<:IO`. `Progress` therefore becomes an
  abstract type; dispatch on `::Progress` and field access are unaffected.
  Code that relied on `Progress` being concrete (e.g. `Vector{Progress}`)
  now gets an abstract eltype.
- `dt` is stored as a `Float64`, converted from any `Real` in the constructor.
- `print_repeat` replaces `"\n" ^ n`; `barstring` builds the bar in an
  `IOBuffer` instead of calling `repeat`.
- `printover` calls a new `printcolored`, which emits exactly the escape codes
  `printstyled` would: only when the output advertises `:color`, looked up in
  `Base.text_colors`, applied per line, skipping empty lines. A test compares
  its output against `printstyled` byte for byte.
- `compute_front` binds `barglyphs.front` to a local once, so it infers to
  `Char`. The public field type is unchanged.

No behavior changes for interactive use. The existing suite passes, plus new
tests for the output type parameters, `dt`, color gating, and `compute_front`
inference.

## Verification

MWE: a package whose `main` constructs `Progress(3; output=Core.stderr, dt=0.0)`,
calls `update!` three times and `finish!`, built with

```
juliac --output-exe pmtrim --trim=safe --bundle build .
```

on Julia 1.13.0 with JuliaC 0.3.10.

| ProgressMeter | verifier errors |
|---|---|
| master (b631816) | 63 |
| this PR | 25 |

None of the remaining 25 are in code this PR touches:

- 9 are Base's `errormonitor` closure, pulled in by `using Distributed`;
  #366 removes them (verified: a package that only does `using ProgressMeter`
  builds clean on #366's branch and fails with exactly these 9 on master).
- 16 are the `Main.IJulia` reflection in `printover`,
  `move_cursor_up_while_clearing_lines` and `running_ijulia_kernel`. #368 moves
  those into an IJulia package extension.

With this PR, #366 and #368 merged together locally, the MWE builds
with 0 verifier errors and the binary runs. This PR and #368 touch adjacent lines
in `printover`, so whichever lands second needs a trivial rebase.

Related: #332.


bjarthur:
@leostenzel

bjarthur:
for those that want to test juliac, i have merged PRs 366, 367, and 368 here: https://github.com/bjarthur/ProgressMeter.jl/tree/bja/integration

bjarthur:
maybe these CI failures on nightly will pass with the changes to #366 ?

src/ProgressMeter.jl
```diff
@@ -66,26 +66,49 @@ end
...
+        enabled::Bool = true,
+        offset::Integer = 0,
+        output::O = stderr,
+        showspeed::Bool = false,
+        kwargs...) where O<:IO
+    ProgressCore{O}(
+        color, desc, Float64(dt), enabled, Int(offset), output, showspeed,
+        1, 0, Threads.ReentrantLock(), 0, 1, false,
+        2*(Threads.nthreads()>1), Threads.threadid(), time(), time(), time())
 end
```
MarcMush: - why `kwargs...` ?
- `safe_lock` needs to be changeable (it is tested in the tests and currently fails)
- probably also `check_iterations` and `lock` (for locking multiple progressbars with the same lock)

bjarthur: all fixed!  thanks for the review.

src/ProgressMeter.jl
```diff
@@ -631,31 +659,46 @@ function printover(io::IO, s::AbstractString, color::Symbol = :color_normal)
...
-                         compute_front(barglyphs, frac_solid),
-                         repeat(string(barglyphs.empty), max(0, nempty)),
-                         barglyphs.rightend)
+    barlen <= 0 && return ""
+    res = IOBuffer()
+    print(res, barglyphs.leftend)
+    if percentage_complete >= 100
+        for _ in 1:Int(barlen)
+            print(res, barglyphs.fill)
+        end
```
MarcMush: Is there a reason for not using your new print_repeat?

MarcMush:
I've only tested on windows where printing is quite slow, and even if there are less allocations, performance is worse on my machine

results from the performance tests in `core.jl`:

```
Performance results: (master)
without progress:      1.13 ns/it
with no lock:         20.00 ns/it
with no printing:     20.16 ns/it
with disabled:         2.26 ns/it
with lock:            47.31 ns/it
with automatic lock:  18.78 ns/it
with lock, disabled:  28.99 ns/it
with force:            0.18 ms/it

Performance results: (this pr)
without progress:      1.13 ns/it
with no lock:         38.15 ns/it
with no printing:     37.92 ns/it
with disabled:        24.31 ns/it
with lock:            66.14 ns/it
with automatic lock:  40.15 ns/it
with lock, disabled:  50.02 ns/it
with force:            0.22 ms/it
```

I'm not even sure it's the printing the problem, because with force the result are very similar, but with disabled, that's where it's the most obvious even if nothing is printed

src/ProgressMeter.jl
bjarthur: doh, yea, here and three other places

bjarthur:
locally one test is failing.  fingers crossed it works on github CI:

```
Core: Test Failed at /Users/arthurb/ProgressMeter.jl/test/core.jl:126
  Expression: t_prog < 9t_noprog
   Evaluated: 2.823451625e-8 < 9.95545872e-9
```

MarcMush:
yes, this is the performance test, is it deactivated in CI because performance is not reliable. It fails on windows because printing is slow

It is also quite arbitrary, because the workload in the function is so low (just a rand())

bjarthur:
f13d9f8 fixes a performance regression this PR had introduced in the non-printing hot path.

**The problem:** parametrizing the meters on the output type made the default constructors type-unstable. `stderr` is a typed global of *abstract* type `IO`, so when `output` is defaulted, `typeof(stderr)` is unknown to inference and `Progress(n)` returns the abstract `Progress{O} where O`. Every subsequent `next!`/`update!` on it then goes through runtime dispatch, adding a near-constant ~20 ns to every iteration in `test/core.jl`'s perf results — including the `enabled=false` case (2.3 → 24.3 ns/it), which is the signature of per-call dynamic dispatch rather than the work itself getting slower.

**The fix:** when no `output` is passed, the constructors now build `ProgressCore{IO}` explicitly (storing `stderr` behind `IO`) instead of parametrizing on its runtime type. `Progress{IO}` is a concrete type, so construction infers concretely and the hot path dispatches statically again, with a dynamic dispatch only at an actual print — exactly master's behavior and layout. An explicitly passed output still yields the fully static `Progress{typeof(output)}`, which is what a `--trim=safe` binary needs anyway (trimmed code cannot rely on the untyped `stderr` global). The dispatch is on a compile-time-constant `nothing` default, so both branches constant-fold.

**One caveat:** writing `Progress(n; output=stderr)` *explicitly* still yields the abstract `Progress{O} where O` at the call site, because `typeof(stderr)` is unknowable at compile time. That is inherent to any design that parametrizes on the output type; such callers can recover master's performance by omitting the kwarg (getting `Progress{IO}`) or keep it behind a function barrier.

Perf on the same machine (Julia 1.13, `-t 4`):

| case | master | PR before fix | PR after fix |
|---|---|---|---|
| no lock | 16.7 ns/it | 38.2 | 15.5 |
| no printing | 15.6 | 37.9 | 16.4 |
| disabled | 2.6 | 24.3 | 2.6 |
| lock | 31.0 | 66.1 | 31.1 |
| automatic lock | 15.8 | 40.2 | 16.5 |

The `--trim=safe` MWE from the description still builds identically: 16 verifier errors with byte-identical signatures before and after the fix, all of them the `Main.IJulia` reflection that #368 removes. New tests assert `@inferred(Progress(10)) isa Progress{IO}` (and likewise for `ProgressThresh`/`ProgressUnknown`, plus the explicit-output parametrized case).

bjarthur:
One more note, unrelated to the fix above: the assertion `t_prog < 9*t_noprog` at `test/core.jl:126` fails on master too on the machine used for the numbers above (16.7 vs 9×1.1 ns) — the bare no-progress loop is simply too fast there, so a 9× multiple of it is under the fixed ~16 ns cost of a `next!` call. It is guarded by `ENV["CI"]`, so CI never sees it (the full suite was run with `CI=true` here), but locally it aborts `include("test/core.jl")` right after the perf results print. It may be worth loosening that threshold, or comparing against an absolute per-iteration budget instead — separately from this PR.

MarcMush:
your LLM-output is starting to get annoying, your "one more note" comment is just repeating what I said previously

are these numbers from tests on windows, linux or mac?

otherwise this all looks good

bjarthur:
M2 mac

bjarthur:
the LLM suggestion was to increase the current factor of 9 so it would pass without having to disable it in CI.  in a separate PR.

MarcMush:
Thank you, a simple phrase from a human is much more nicer and actually feels clearer to me

It currently doesn't make sense: if in a new version, Julia improves the performance of rand(), it would appear that we're doing worse, which is also not right

A fixed time would then fail on a non-powerful machine, and especially in github CI, performance isn't reliable, probably depending on their current server load and because this is a free tier

A way to detect regression compared to master would be great (like we did here manually)

(quick Google search later) maybe something like that? https://discourse.julialang.org/t/how-to-implement-performance-regression-tests/132578/3

PR #368, opened 2026-09-15
Move IJulia support to a package extension

bjarthur:
ProgressMeter reaches IJulia by reflection on `Main`:
`isdefined(Main, :IJulia)`, `Main.IJulia.inited`, `Main.IJulia.clear_output(true)`,
and the `stdio_bytes` / `reset_stdio_count` workaround for #76. Every one of
those is a dynamic call, so a trimmed binary (`juliac --trim=safe`) that
updates a meter fails the verifier on them even though no notebook is involved.

## Changes

- An `IJuliaBackend` abstract type with an inert `NoIJulia` default stored in
  `ProgressMeter.IJULIA_BACKEND`. Four hooks dispatch on it:
  `ijulia_loaded`, `ijulia_running`, `ijulia_clear_output`, `ijulia_reset_stdio`.
  In the base package each has exactly one method, so the calls resolve
  statically.
- `ext/ProgressMeterIJuliaExt.jl`, loaded by `using IJulia`, defines a
  `WithIJulia` backend, adds the real methods, and registers it in `__init__`.
  The `pkgversion(IJulia) < v"1.30"` switch for #76 moves there unchanged.
- `running_ijulia_kernel`, `move_cursor_up_while_clearing_lines` and `printover`
  call the hooks instead of touching `Main`. No more `Main.IJulia` anywhere.
- README: one line noting the integration is an extension.

## Behavior

With IJulia loaded, behavior matches master. One deliberate widening: the
previous `isdefined(Main, :IJulia)` was true only if IJulia was imported into
`Main`; the extension loads whenever IJulia is loaded anywhere in the session.
In a running kernel both hold, so notebook users see no difference.

## Tests

- Without IJulia: the default backend is `NoIJulia`, `running_ijulia_kernel()`
  is `false`, the hooks return `nothing`, and `printover` still emits the
  line clear.
- With IJulia (run manually, since adding IJulia to the test target would pull
  ZMQ and a kernelspec install into CI): the extension loads, the backend is
  `WithIJulia`, `ijulia_loaded` is `true`, `running_ijulia_kernel()` is `false`
  outside a kernel, and `printover` no longer emits the line clear.

## Verification under --trim

MWE and method as in the companion type-parametrization PR #367.

| ProgressMeter | verifier errors |
|---|---|
| master (b631816) | 63 |
| this PR | 59 |

All `Main.IJulia`-related errors are gone. The remaining 59 are in code this
PR does not touch: 50 dynamic dispatches on `ProgressCore.output::IO`,
`dt::Real`, `repeat` and `printstyled`, fixed by #367, and 9 from
`using Distributed`, fixed by #366. With all three merged together locally,
the MWE builds with 0 verifier errors and the binary runs. This PR and the
companion touch adjacent lines in `printover`, so whichever lands second needs
a trivial rebase.

Related: #332, #366.


bjarthur:
@leostenzel

bjarthur:
for those that want to test juliac, i have merged PRs 366, 367, and 368 here:  https://github.com/bjarthur/ProgressMeter.jl/tree/bja/integration

